PR #2, opened 2019-06-04
Fix _ps_const_ty macro compile on non-x86

repi:
Was building for `wasm32-unknown-unknown` when I ran into this, but should be a fail on any non-x86 target. But easy fix!


bitshifter:
Thanks!

PR #8, opened 2019-07-22
Method documentation for `Vec4Mask`.

cessen:
Documented all methods on `Vec4Mask`.

### DIFF
--- a/src/f32/vec4_f32.rs
+++ b/src/f32/vec4_f32.rs
@@ -550,6 +550,10 @@ impl Distribution<Vec4> for Standard {
     }
 }
 
+/// A 4-dimensional vector mask.
+///
+/// This type is typically created by comparison methods on `Vec4`.  It is
+/// essentially a vector of four boolean values.
 #[derive(Clone, Copy, Default)]
 // if compiling with simd enabled assume alignment needs to match the simd type
 #[cfg_attr(not(feature = "scalar-math"), repr(align(16)))]
@@ -557,6 +561,7 @@ impl Distribution<Vec4> for Standard {
 pub struct Vec4Mask(u32, u32, u32, u32);
 
 impl Vec4Mask {
+    /// Creates a new `Vec4Mask`.
     pub(crate) fn new(x: bool, y: bool, z: bool, w: bool) -> Self {
         const MASK: [u32; 2] = [0, 0xff_ff_ff_ff];
         Self(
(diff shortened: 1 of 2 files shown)


src/f32/vec4_f32.rs
```diff
@@ -550,13 +550,18 @@ impl Distribution<Vec4> for Standard {
     }
 }
 
+/// A 4-dimensional vector mask.
+///
+/// This type is typically created by comparison methods on `Vec4`.  It is
+/// essentially a vector of four boolean values.
```
bitshifter: It might be worth mentioning that it's stored using four `u32` values. True is always `0xffffffff` and false is `0x0`. The scalar implementation also uses `u32` also so that size and alignment and values will match the SIMD version.

src/f32/vec4_f32.rs
```diff
@@ -573,21 +578,38 @@ impl Vec4Mask {
         self.bitmask()
     }
 
+    /// Returns a bitmask with the lowest four bits set from the elements of
```
bitshifter: It might be worth mentioning that only the most significant bit of each lane (element) is used for creating the bitmask. I'm using the packed_simd docs for inspiration here https://rust-lang-nursery.github.io/packed_simd/packed_simd/type.m32x4.html#method.bitmask

src/f32/vec4_f32.rs
cessen: Are those semantics part of the API promise here?  If so, then I agree we should add it.  But I don't see where it's exposed at the moment.

src/f32/vec4_f32.rs
cessen: Same question here as above.

cessen: I guess to expand on that: is this crate planning to only support x86/64 SIMD into the future?  I expect that specifics like this would vary from platform to platform.

src/f32/vec4_f32.rs
bitshifter: I suppose the fact that internally `0xffffffff` is true isn't part of the public API. The size and the alignment is part of the public interface though.

src/f32/vec4_f32.rs
bitshifter: I'd like to support other architectures at some point. I'm working on the assumption that this holds across different architectures because `packed_simd` supports multiple architectures. That said, `packed_simd` is an experimental crate and might not be correct.

bitshifter:
Merged this because ultimately I'm happy with it and I figure it's easier to make any tweaks later, plus it will probably make your life easier given you've got subsequent changes. Thanks!

cessen:
Thanks much!

Yeah, I'm more than happy to make further tweaks.

PR #12, opened 2019-07-22
Added basic logic operations for the vector mask types.

cessen:
Implemented bitwise `and`, `or`, and `not` operators for `Vec4Mask`, `Vec3Mask`, and `Vec2Mask`.

### DIFF
--- a/src/f32/vec2.rs
+++ b/src/f32/vec2.rs
@@ -402,3 +402,39 @@ impl Vec2Mask {
         )
     }
 }
+
+impl BitAnd for Vec2Mask {
+    type Output = Vec2Mask;
+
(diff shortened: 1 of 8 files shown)


cessen:
Hmm, it also occurs to me that the logic+assignment operators would be handy.  Hold on before you merge this, and I'll implement those as well.

cessen:
Okay, added the assignment operators too!  If it looks good to you, please go ahead and merge. :-)

And thank you for being so open to my contributions!

bitshifter:
Thanks for the PRs! This is all functionality I wanted to add but hadn't got around to it so I'm happy to take the changes.

PR #16, opened 2019-07-23
Make `Vec4::dup_*` methods public.

cessen:


cessen:
On second thought, I think maybe we shouldn't expose these just yet.

Issue #6, opened 2019-07-20
Remove approx and rand from default features


bitshifter:
approx and rand define traits which must be implemented inside glam for them to be used in test and bench executables. Unfortunately this means they must be enabled by default so that a simple `cargo test` or `cargo bench` will work without needing to specify additional features. I'd rather not enable them by default.

For the purposes of glam's tests and benches it is probably easier to implement a simple non-crypto rand and some approx eq functions.

`glam` could use it's own `approx_eq` functions anyway as we don't re-export the `approx` crate, I'd prefer not to need users to have to import traits for this and the `approx` traits aren't documented much.

The `rand` crate is quite complex and we're not concerned with crypto strength random numbers for the purposes of our unit tests.

The current implementations of `approx` and `rand` can be kept unless it becomes a burden to maintain them.

- [ ] approx
- [x] rand

cessen:
I responded to your comment on my commit to Psychopath already, but reading this issue I realized I misunderstood a bit.

I _think_ you don't actually need to have these enabled by default for using them in testing.  You can just have them be a dev dependency, and have the impls hidden behind a `#[cfg(test)]` attribute.  Maybe there's corner-cases I'm not aware of, but I think that works.

If that doesn't work, then rolling your own and making the testing-only definitely would, and I'd be in favor of that.  I don't actually need approximate comparisons implemented in glam--I'm happy to roll my own as needed, and it seems orthogonal to glam's purpose.

repi:
Would also like to see both of them disabled by default, and ideally replacing `approx::assert_ulps_eq` in the tests with something internal in the crate to avoid always pulling in `approx`.

PR #19, opened 2019-09-14
Update rand 0.7 & criterion 0.4 and disable OS random number generator

repi:
Disabled default features on `rand` to avoid pulling in things that are not needed, such as the OS random number generator in `getrandom`. 

Want to be able to build for `wasm32-unknown-unknown` without having this pulled in because if called it will just panic (as there is no OS).

Unfortunately I think in practice due to `criterion` using `rand` with all features it will enable all features here also even though `criterion` is just a dev-dependency. But that is a separate cargo issue for now.

And good to use latest `rand` 0.7 anyway with minimal features.

And I do agree with #6 that ideally both `rand` and `approx` should be disabled by default.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -38,7 +38,7 @@ rand = { version = "0.7", optional = true, default-features = false }
 serde = { version = "1.0", optional = true, features = ["derive"] }
 
 [dev-dependencies]
-criterion = "0.2"
+criterion = "0.3"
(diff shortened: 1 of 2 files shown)


bitshifter:
Thanks again. I'd like to make the rand dependency optional at some point. I just need to remove the dependency on it in tests and bench.

PR #20, opened 2019-09-17
Fix Mat4::look_at_lh/rh

repi:
The `Mat4::look_at_lh` implementation was a broken with lacking normalization, doing the cross products in the wrong order and some flips. 

I'm not 100% sure if this is correct but with these fixes it matches the behavior of `glm::look_at_rh` and also the implementation in DirectXMath `XMMatrixLookToLH`, though I've only in practice tested the right-handed version as that is what we use.

### DIFF
--- a/src/f32/mat4.rs
+++ b/src/f32/mat4.rs
@@ -390,7 +390,7 @@ impl Mat4 {
     fn look_to_lh(eye: Vec3, dir: Vec3, up: Vec3) -> Self {
         let f = dir.normalize();
         let s = up.cross(f).normalize();
-        let u = f.cross(s).normalize();
+        let u = f.cross(s);
         let (fx, fy, fz) = f.into();
         let (sx, sy, sz) = s.into();
         let (ux, uy, uz) = u.into();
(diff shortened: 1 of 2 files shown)


src/f32/mat4.rs
```diff
@@ -389,17 +389,17 @@ impl Mat4 {
     // TODO: make public at some point
     fn look_to_lh(eye: Vec3, dir: Vec3, up: Vec3) -> Self {
         let f = dir.normalize();
+        let s = up.cross(f).normalize();
+        let u = f.cross(s).normalize();
```
MaikKlein: You can get rid of that normalize. f and s are already normalized.

repi: Yup good point, will fix

bitshifter:
Thanks. I don't use these in my little hobby projects so they haven't really had any love. I think I copied them from nalgebra or something but obviously messed it up.

They're still not public though, are you using them somehow or did you just notice they were incorrect?

repi:
Yeah easy to get something wrong with it. The `Mat4::look_to_lh` is not public but `Mat4::look_to_lh` and `Mat4::look_to_rh` that uses it are public.  And we use `Mat4::look_at_rh`.

Thanks for quick merge!

PR #23, opened 2019-09-17
Add support for infinite projection matrices

repi:
Both standard and reverse depth.

The infinite reverse projection matrix is the primary one we use, and it is the one with the best precision and the simplest construction. So really the one most everyone should use by default.

Difficult to unit test this, but multiple of us have verified the construction of it.

Also probably should rename the `perspective_glu` function here also, or maybe (depending on how opinionated we want to be) simply remove it? It is only useful in OpenGL due to the -1,1 depth range and it should have a `_rh` suffix to indicate it is right-handed.

### DIFF
--- a/src/f32/mat4.rs
+++ b/src/f32/mat4.rs
@@ -415,14 +415,15 @@ impl Mat4 {
     }
 
     #[inline]
-    /// Returns the equivalent of the common pespective function `gluPerspective`.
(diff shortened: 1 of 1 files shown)


bitshifter:
Thanks. The main driver for unit tests is ensuring the SIMD and scalar code behaves the same, so shouldn't matter too much here, other than keeping code coverage happy...

PR #31, opened 2019-11-06
Mat4::from_scale(Vec3): Do not assert any zero scale component, only on all zero

Matthias-Fauconneau:
I think default panic on any zero scale component is too opiniated.
It can be valid to squash along an axis, e.g for UI.
For the Z special case, from_scale(Vec2) could be defined, with scale.z = 0 or 1.

P.S: Feel free to merge without history
(Triggered by iced/wgpu:hidpi-support)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 **/*.rs.bk
 Cargo.lock
 Session.vim
+.cargo-ok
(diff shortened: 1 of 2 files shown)


Matthias-Fauconneau:
Sounds like a good idea :)

I had first tried to run iced with glam as a cargo git dep and then moved
the working copy out of cargo cache to edit it.
Cargo seems to touch this file to tell whether it checked something out.
Could have just removed it.
I guess you probably generally won't want to commit hidden files, so you
could just ignore .*

PR #35, opened 2019-12-15
Add cargo-deny GitHub CI action

repi:
To verify all dependencies are using permissible licenses and that there are no duplicate versions of dependencies.

For a small crate like this it is likely not super useful, but hoping it could help a bit and is easy to set up and use.

For more info: 
https://github.com/EmbarkStudios/cargo-deny
https://github.com/EmbarkStudios/cargo-deny-action

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -1,0 +1,8 @@
+name: CI
+on: [push, pull_request]
+jobs:
+  cargo-deny:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v1
+      - uses: EmbarkStudios/cargo-deny-action@v0
--- a/deny.toml
+++ b/deny.toml
@@ -1,0 +1,20 @@
+[bans]
+multiple-versions = "deny"
+deny = []
+skip = [
+    # criterion uses old 0.3, glam uses 0.4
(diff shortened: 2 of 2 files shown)


repi:
Ah that didn't go that well, GitHub Actions not enabled/supported in this repo?

Thought about moving over from Travis to GitHub Actions? We've done it for all of our repos and have been working really well and quite a bit faster.

If it is enabled in this repo one should be able to continue use travis for the main build and GitHub Action for this check though

bitshifter:
Local and third party Actions appear to be enabled for this repository. I haven't used Actions at all. What went wrong?

repi:
Not fully sure, usually it picks up the GitHub Action directly when adding the file for it and starts build  it as an additional check in the repo. AFAIK there is no other configuration for enabling a GitHub Action, should be enough to have the `.github/workflows/xxx` files

repi:
But in this PR it didn't trigger that new workflow / check

bitshifter:
Maybe it needs to be merged before it will work? I don't see any harm in merging it in any case.

repi:
Yeah could be that it needs to be on master first if the PR creator is not the repo owner.

Can be merged, then one can at a minimum run it locally with `cargo deny check` in the root. But hopefully the CI action gets going also

repi:
That worked!

PR #38, opened 2019-12-20
Added Mat4::perspective_lh_dx + minor consistency changes.

michal-z:


michal-z:
Yes, DX, Vulkan and Metal all use zero-one range. So, maybe just perspective_lh (perspective_infinite_lh, perspective_infinite_reverse_lh)?

bitshifter:
Sounds good to me!

PR #40, opened 2019-12-21
Added Mat4::perspective_infinite_lh and Mat4::perspective_infinite_re…

michal-z:
…verse_lh.


bitshifter:
Thanks! Will merge it when I'm next at my PC.

PR #42, opened 2019-12-26
Added Vec2::round(), Vec3::round() and Vec4::round() (scalar and sse2 versions).

michal-z:
Added Vec2::round(), Vec3::round() and Vec4::round() (scalar and sse2 versions).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog], and this project adheres to
 
 ## [Unreleased]
 
+### Added
+* Added `Mat4::perspective_lh`, `Mat4::perspective_infinite_lh` and
+  `Mat4::perspective_infinite_reverse_lh`.
+
 ## [0.8.4] - 2019-12-17
 
 ### Added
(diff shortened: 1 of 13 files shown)


bitshifter:
Away from PC at the moment, looks like you've addressed most things I was going to raise already. In always interested in the use case for new features and how people are using the library, what algorithm are you using `round` in?

michal-z:
I'm very interested in pushing the graphics/games development Rust ecosystem forward. I think we need fast and rich math library like DirectXMath. So, I would like to port some features from this C++ lib. Also, I'm working on some creative graphics experiments using Rust and D3D12.  I want to push a lot of preprocessed data from CPU to GPU. I need fast round()/floor()/ceil() for noise-based effects and also sin/cos. In the next PRs I would like to implement vectorized versions of sin/cos/tan if you are interested.

bitshifter:
Yep that all makes sense. I did have an sse2 sin_cos that is disabled for now as it lost accuracy for large angle values, so that's probably something that would need to be handled. Or if accuracy was a lot different from std library trig functions it might be better to make them opt in.

You might want to create a new branch from master for future PRs once this is merged since git is including a lot of changes I've already merged in the PR. I have been squashing commits which might be confusing history on your branch.

PR #43, opened 2019-12-28
Added VecN::floor and VecN::ceil (sse2 and scalr versions).

michal-z:


michal-z:
Those functions handle INF and -INF just fine. NANs are handled differently than in scalar versions. SIMD versions still produce NANs but with different bit encoding.
So, this test fails:
assert_eq!(new::Vec3(f32::NAN, 0.0, 0.0).floor().x(), f32::NAN);

But this one passes:
assert_eq!(new::Vec3(f32::NAN, 0.0, 0.0).floor().x().is_nan(), f32::NAN.is_nan());

We still get NANs (all exponent bits are '1') but with different mantissa bits. I think it can be tricky to implement scalar versions that behaves the same.

Personally I think that we should allow this for performance reasons. What do you think?

bitshifter:
It's fine if the bit pattern is different, as long as `is_nan` is consistent. Thanks for checking.

bitshifter:
Did you want to add any additional test commits or just leave it?

michal-z:
I will add test later today.

PR #44, opened 2020-01-17
Add missing type docs for Vec3, Vec3Mask & Vec4

repi:


repi:
Cool, and yeah think that would be better longer term solution

bitshifter:
I can't merge this on my phone because of the CI fail so I'll have to do it later when I'm at my PC again.

PR #49, opened 2020-03-03
Added x,y,z,w methods to Quat

Chris-Nicholls:
This adds methods to read the elements of a quaternion directly. 
You can currently do this by using the various `From` traits implemented on Quat, e.g.: 
```rust
let (x, y, z, w) = quat;
```
or 

```rust
let x = Vec4::from(quat).x();
```
The first of these has the disadvantage that you don't know for sure that you got the order correct without looking up the source. The second is better, perhaps, but again, it's not obvious that it's valid without checking, since the Quat type could well store its elements in the order `w`,`x`,`y`,`z`.

Adding these methods removes the need for any indirection, brings Quat more in line with other data types, and I think it leads to more obviously correct code.

### DIFF
--- a/src/f32/quat.rs
+++ b/src/f32/quat.rs
@@ -441,6 +441,29 @@ impl Quat {
             Self(Vec4(_mm_add_ps(result0, result1)))
         }
     }
+    /// Returns element `x`.
+    #[inline]
+    pub fn x(self) -> f32 {
+        self.0.x()
+    }
+
+    /// Returns element `y`.
+    #[inline]
+    pub fn y(self) -> f32 {
+        self.0.y()
+    }
+
+    /// Returns element `z`.
+    #[inline]
+    pub fn z(self) -> f32 {
+        self.0.z()
+    }
+
+    /// Returns element `w`.
+    #[inline]
+    pub fn w(self) -> f32 {
+        self.0.w()
+    }
 }
 
 impl fmt::Debug for Quat {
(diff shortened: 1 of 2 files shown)


bitshifter:
Hi @Chris-Nicholls , thanks for the PR. The main reason I didn't add accessors to `Quat` was I didn't know of a use case where people might want to access the individual elements. What is your use case for adding these?

Chris-Nicholls:
Hey,
So my use case is serialising the quaternions into a binary format for a
multiplayer game. You rarely want to sync four floats for quaternions, and
instead want to compress them in smallest threes or quantize a la
https://gafferongames.com/post/snapshot_compression/

For this you need to access the individual x,y,z,w elements.

bitshifter:
Makes sense. It is worth keeping in mind that because `Quat` is stored in SIMD vector it is slightly more expensive to access each element individually. As you point out `From` could be ambiguous. On the matrix types I added more explicit `to` methods (e.g. `to_cols_array`) for the same reason. Do you think it would make sense to add a `to_xyzw` here? This would mirror the `from_xyzw` method that's already there.

I can merge this PR regardless, but was interested if you thought that might be useful.

Chris-Nicholls:
I like the idea of `to_xyzw` but, at least for my use case, it's very unlikely to provide a noticeable performance difference since the rest of the serialization will dominate.

PR #52, opened 2020-04-14
Run cargo test in CI

aloucks:


aloucks:
Not sure how the unit test coverage could drop when there was no addition or modification of code!

bitshifter:
cargo build and test are run by travis. You can see all the builds for your pull requests here https://travis-ci.org/github/bitshifter/glam-rs/pull_requests

I am not sure why they aren't showing up in the list of checks though.

bitshifter:
I'll leave CI on travis for now, since I think it's probably a larger job to reach feature parity with the travis config. I will try and find out why it's not showing up in the checks though.

PR #57, opened 2020-04-17
Add Quat::slerp

aloucks:
This PR adds `Quat::slerp`. This implementation is derived from the cgmath implementation which is based on http://number-none.com/product/Understanding%20Slerp,%20Then%20Not%20Using%20It/

Unlike the cgmath implementation , intrinsics are used here when available.

Opening as draft to gain feedback.

1. `_MM_SHUFFLE` is considerably more readable than the duplication of the binary constants. Although it's implemented in the rust std library, it's currently nightly only. I've copied that implementation here.
2. I've added a few macros that mirror the DirectXMath macros, but I'm not sure if these are something you'll want in the code.
3. I've used the DirectXMath constant names which aren't snake case. I think it makes it easier when comparing the code, but I can understand if you wouldn't want that.

### DIFF
--- a/src/f32/funcs.rs
+++ b/src/f32/funcs.rs
@@ -115,20 +115,6 @@ pub(crate) mod sse2 {
     _ps_const_ty!(PS_TWO_PI, f32x4, std::f32::consts::PI * 2.0);
     _ps_const_ty!(PS_RECIPROCAL_TWO_PI, f32x4, 0.159154943);
 
-    #[cfg(target_feature = "avx")]
-    macro_rules! m128_permute {
-        ($v:expr, $c:expr) => {
-            _mm_permute_ps($v, $c)
-        };
-    }
-
-    #[cfg(not(target_feature = "avx"))]
-    macro_rules! m128_permute {
-        ($v:expr, $c:expr) => {
-            _mm_shuffle_ps($v, $v, $c)
-        };
-    }
-
     #[cfg(target_feature = "fma")]
     macro_rules! m128_mul_add {
         ($a:expr, $b:expr, $c:expr) => {
@@ -248,19 +234,19 @@ pub(crate) mod sse2 {
 
         // Compute polynomial approximation
         const SC1: __m128 = unsafe { PS_SIN_COEFFICIENTS1.m128 };
-        let v_constants_b = m128_permute!(SC1, 0b00_00_00_00);
+        let v_constants_b = _mm_shuffle_ps(SC1, SC1, 0b00_00_00_00);
 
         const SC0: __m128 = unsafe { PS_SIN_COEFFICIENTS0.m128 };
-        let mut v_constants = m128_permute!(SC0, 0b11_11_11_11);
+        let mut v_constants = _mm_shuffle_ps(SC0, SC0, 0b11_11_11_11);
         let mut result = m128_mul_add!(v_constants_b, x2, v_constants);
 
-        v_constants = m128_permute!(SC0, 0b10_10_10_10);
+        v_constants = _mm_shuffle_ps(SC0, SC0, 0b10_10_10_10);
         result = m128_mul_add!(result, x2, v_constants);
 
-        v_constants = m128_permute!(SC0, 0b01_01_01_01);
+        v_constants = _mm_shuffle_ps(SC0, SC0, 0b01_01_01_01);
         result = m128_mul_add!(result, x2, v_constants);
 
-        v_constants = m128_permute!(SC0, 0b00_00_00_00);
+        v_constants = _mm_shuffle_ps(SC0, SC0, 0b00_00_00_00);
         result = m128_mul_add!(result, x2, v_constants);
 
         result = m128_mul_add!(result, x2, PS_ONE.m128);
--- a/src/f32/funcs.rs
+++ b/src/f32/funcs.rs
@@ -60,6 +60,13 @@ pub(crate) mod sse2 {
                 $field: [$x, $x, $x, $x],
             };
         };
+
+        ($name:ident, $field:ident, $x:expr, $y:expr, $z:expr, $w:expr) => {
+            #[cfg(any(target_arch = "x86", target_arch = "x86_64"))]
+            const $name: UnionCast = UnionCast {
+                $field: [$x, $y, $z, $w],
+            };
+        };
     }
 
     _ps_const_ty!(PS_INV_SIGN_MASK, u32x4, !0x8000_0000);
@@ -84,6 +91,28 @@ pub(crate) mod sse2 {
     // _ps_const_ty!(PS_COSCOF_P1, f32x4, -1.388_731_6E-3);
     // _ps_const_ty!(PS_COSCOF_P2, f32x4, 4.166_664_6e-2);
     // _ps_const_ty!(PS_CEPHES_FOPI, f32x4, 1.273_239_5); // 4 / M_PI
+
+    _ps_const_ty!(g_XMNegativeZero, u32x4, 0x80000000);
+    _ps_const_ty!(g_XMPi, f32x4, std::f32::consts::PI);
+    _ps_const_ty!(g_XMHalfPi, f32x4, std::f32::consts::FRAC_PI_2);
+    _ps_const_ty!(g_XMSinCoefficients0 , f32x4, -0.16666667, 0.0083333310, -0.00019840874, 2.7525562e-06);
+    _ps_const_ty!(g_XMSinCoefficients1 , f32x4, -2.3889859e-08, -0.16665852 /*Est1*/, 0.0083139502 /*Est2*/, -0.00018524670 /*Est3*/);
+    _ps_const_ty!(g_XMOne, f32x4, 1.0);
+    _ps_const_ty!(g_XMTwoPi, f32x4, std::f32::consts::PI * 2.0);
+    _ps_const_ty!(g_XMReciprocalTwoPi , f32x4, 0.159154943);
+
+    macro_rules! XM_PERMUTE_PS {
+        ($v:expr, $c:expr) => {
+            _mm_shuffle_ps($v, $v, $c)
+        }
+    }
+
+    // XMVectorNegativeMultiplySubtract
+    macro_rules! XM_FMADD_PS {
+        ($a:expr, $b:expr, $c:expr) => {
+            _mm_fmadd_ps($a, $b, $c)
+        }
+    }
 
     #[inline]
     pub(crate) unsafe fn m128_round(v: __m128) -> __m128 {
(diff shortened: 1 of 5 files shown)


src/f32/funcs.rs
```diff
@@ -85,6 +92,28 @@ pub(crate) mod sse2 {
...
     // _ps_const_ty!(PS_CEPHES_FOPI, f32x4, 1.273_239_5); // 4 / M_PI
 
+    _ps_const_ty!(g_XMNegativeZero, u32x4, 0x80000000);
+    _ps_const_ty!(g_XMPi, f32x4, std::f32::consts::PI);
+    _ps_const_ty!(g_XMHalfPi, f32x4, std::f32::consts::FRAC_PI_2);
+    _ps_const_ty!(g_XMSinCoefficients0 , f32x4, -0.16666667, 0.0083333310, -0.00019840874, 2.7525562e-06);
+    _ps_const_ty!(g_XMSinCoefficients1 , f32x4, -2.3889859e-08, -0.16665852 /*Est1*/, 0.0083139502 /*Est2*/, -0.00018524670 /*Est3*/);
+    _ps_const_ty!(g_XMOne, f32x4, 1.0);
+    _ps_const_ty!(g_XMTwoPi, f32x4, std::f32::consts::PI * 2.0);
+    _ps_const_ty!(g_XMReciprocalTwoPi , f32x4, 0.159154943);
```
bitshifter, line 102: I would want these to use Rust naming conventions.

src/f32/funcs.rs
```diff
@@ -85,6 +92,28 @@ pub(crate) mod sse2 {
...
+    macro_rules! XM_PERMUTE_PS {
+        ($v:expr, $c:expr) => {
+            _mm_shuffle_ps($v, $v, $c)
+        }
+    }
+
+    // XMVectorNegativeMultiplySubtract
+    macro_rules! XM_FMADD_PS {
+        ($a:expr, $b:expr, $c:expr) => {
+            _mm_fmadd_ps($a, $b, $c)
```
bitshifter, line 113: Need to check the fma target feature is available to use this.

src/f32/funcs.rs
```diff
@@ -85,6 +92,28 @@ pub(crate) mod sse2 {
...
+    _ps_const_ty!(g_XMNegativeZero, u32x4, 0x80000000);
+    _ps_const_ty!(g_XMPi, f32x4, std::f32::consts::PI);
+    _ps_const_ty!(g_XMHalfPi, f32x4, std::f32::consts::FRAC_PI_2);
+    _ps_const_ty!(g_XMSinCoefficients0 , f32x4, -0.16666667, 0.0083333310, -0.00019840874, 2.7525562e-06);
+    _ps_const_ty!(g_XMSinCoefficients1 , f32x4, -2.3889859e-08, -0.16665852 /*Est1*/, 0.0083139502 /*Est2*/, -0.00018524670 /*Est3*/);
+    _ps_const_ty!(g_XMOne, f32x4, 1.0);
+    _ps_const_ty!(g_XMTwoPi, f32x4, std::f32::consts::PI * 2.0);
+    _ps_const_ty!(g_XMReciprocalTwoPi , f32x4, 0.159154943);
+
+    macro_rules! XM_PERMUTE_PS {
```
bitshifter, line 104: I'm not sure it's worth keeping these macros. I think they can be useful when dealing with different implementations of common operations, (e.g. SSE2 and SSE4) but so far I've just stuck with SSE2 and written everything in explicit intrinsices.

src/f32/funcs.rs
```diff
@@ -139,6 +168,69 @@ pub(crate) mod sse2 {
         _mm_or_ps(result, _mm_castsi128_ps(test))
     }
 
+    #[inline]
+    #[allow(non_snake_case)]
+    pub(crate) const fn _MM_SHUFFLE(z: u32, y: u32, x: u32, w: u32) -> i32 {
+        // core::arch::x86_64::_MM_SHUFFLE requires nightly
+        ((z << 6) | (y << 4) | (x << 2) | w) as i32
+    }
```
bitshifter, line 176: I prefer the binary literals. I used to use `_MM_SHUFFLE` but actually found it harder to follow what was going on so migrated to binary literals instead.

src/f32/quat.rs
```diff
@@ -358,6 +358,64 @@ impl Quat {
...
+        
+        const DOT_THRESHOLD: f32 = 0.9995;
+
+        let dot = self.dot(end);
+
+        if dot > DOT_THRESHOLD {
+            // assumes lerp returns a normalized quaternion
+            self.lerp(end, s)
+        } else {
+            #[cfg(not(vec4sse2))]
```
bitshifter, line 381: use `#[cfg(vec4f32)]`, at some point I hope to support other architectures so `#[cfg(not(vec4sse2))]` would be incorrect once that happens.

src/f32/funcs.rs
```diff
@@ -139,6 +168,69 @@ pub(crate) mod sse2 {
...
+    /// 
+    /// Returns a vector whose components are the corresponding components of Angles modulo 2PI.
+    pub(crate) unsafe fn m128_mod_angles(angles: __m128) -> __m128 {
+        let v = _mm_mul_ps(angles, g_XMReciprocalTwoPi.m128);
+        let v = m128_round(v);
+        XM_FMADD_PS!(g_XMTwoPi.m128, v, angles) // XMVectorNegativeMultiplySubtract
+    }
+
+    /// From DirectXMath: XMVectorSin
+    pub(crate) unsafe fn m128_sin(v: __m128) -> __m128 {
```
bitshifter, line 188: Would be good to add a test for this similar to the `scalar_acos` test to get an idea of the error.

src/f32/funcs.rs
```diff
@@ -85,6 +92,28 @@ pub(crate) mod sse2 {
...
+    _ps_const_ty!(g_XMTwoPi, f32x4, std::f32::consts::PI * 2.0);
+    _ps_const_ty!(g_XMReciprocalTwoPi , f32x4, 0.159154943);
+
+    macro_rules! XM_PERMUTE_PS {
+        ($v:expr, $c:expr) => {
+            _mm_shuffle_ps($v, $v, $c)
+        }
+    }
+
+    // XMVectorNegativeMultiplySubtract
```
bitshifter, line 110: This comment looks wrong, this is fused multiply add

src/f32/funcs.rs
```diff
@@ -139,6 +168,69 @@ pub(crate) mod sse2 {
...
     }
 
+    #[inline]
+    #[allow(non_snake_case)]
+    pub(crate) const fn _MM_SHUFFLE(z: u32, y: u32, x: u32, w: u32) -> i32 {
+        // core::arch::x86_64::_MM_SHUFFLE requires nightly
+        ((z << 6) | (y << 4) | (x << 2) | w) as i32
+    }
+
+    /// From DirectXMath: XMVectorModAngles
```
bitshifter, line 178: I'd put this comment inside the function rather than the doc comment.

src/f32/funcs.rs
```diff
@@ -139,6 +168,69 @@ pub(crate) mod sse2 {
...
+    /// 
+    /// Returns a vector whose components are the corresponding components of Angles modulo 2PI.
+    pub(crate) unsafe fn m128_mod_angles(angles: __m128) -> __m128 {
+        let v = _mm_mul_ps(angles, g_XMReciprocalTwoPi.m128);
+        let v = m128_round(v);
+        XM_FMADD_PS!(g_XMTwoPi.m128, v, angles) // XMVectorNegativeMultiplySubtract
+    }
+
+    /// From DirectXMath: XMVectorSin
```
bitshifter, line 187: Same with this comment

src/f32/quat.rs
```diff
@@ -358,6 +358,64 @@ impl Quat {
...
     }
 
+    /// Performs a spherical linear interpolation between `self` and `other`
+    /// based on the value `s`.
+    ///
+    /// When `s` is `0.0`, the result will be equal to `self`.  When `s`
+    /// is `1.0`, the result will be equal to `other`.
+    #[inline]
+    pub fn slerp(self, end: Self, s: f32) -> Self {
+        // http://number-none.com/product/Understanding%20Slerp,%20Then%20Not%20Using%20It/
```
bitshifter, line 368: I was going to add slerp when I wrote glam, read this article and decided not to :) Well, partially that was the reason. I was going for a minimal feature set which was somewhat based on what https://github.com/nfrechette/rtm supported. It also doesn't have a slerp. Are you using slerp for something?

src/f32/funcs.rs
aloucks: I've updated these detect target CPU features.

src/f32/funcs.rs
aloucks: The `fma` instructions seem to add a significant speedup. 

```
+avx ~5%
+fma ~20%
+fma,+avx ~20%
```

src/f32/quat.rs
aloucks: Yes. I was looking at migrating some stuff from cgmath to glam and slerp was missing. I was checking animations with some of the [glTF sample models](https://github.com/KhronosGroup/glTF-Sample-Models) and the lack of constant velocity makes `lerp` unsuitable.

src/f32/funcs.rs
aloucks: I've changed this to use binary literals

src/f32/funcs.rs
aloucks: Hmm, I seem to get different results when running all of the quat tests vs just the quat slerp test. The `fma` instruction is without question an improvement. It's not clear whether the `avx` one is useful or not.

aloucks:
@bitshifter I've addressed all the feedback and added some tests for slerp.

src/f32/quat.rs
bitshifter: Thanks, that's good to know.

src/f32/funcs.rs
bitshifter: Not sure what you mean, the quat slerp test is changing the result of the other tests? Or enabling AVX is changing results? Having FMA off or on could definitely give different results.

src/f32/funcs.rs
```diff
@@ -139,6 +197,64 @@ pub(crate) mod sse2 {
...
     }
 
+    /// Returns a vector whose components are the corresponding components of Angles modulo 2PI.
+    #[allow(dead_code)]
+    pub(crate) unsafe fn m128_mod_angles(angles: __m128) -> __m128 {
+        // From DirectXMath: XMVectorModAngles
+        let v = _mm_mul_ps(angles, XM_RECIPROCAL_TWO_PI.m128);
+        let v = m128_round(v);
+        XM_FMADD_PS!(XM_TWO_PI.m128, v, angles)
+    }
```
bitshifter, line 207: If this is no longer used it should probably be removed. Although I ask about whether this should be used below.

src/f32/funcs.rs
bitshifter: With `XM_PERMUTE_PS` and `XM_FMADD_PS` I'd prefer to make them inline functions and name them `m128_permute` and `m128_mul_add`. Note that `avx` and `fma` features aren't enabled by default in x64 builds and I probably won't add build permutations for these to the CI, just because it already takes an hour as is so they won't be getting tested. It might be worth making a note about that in the README. Well, if I could add a travis build that wasn't waited on by the CI (so PR's weren't waiting on them) but still let me know if they are broken maybe I could add an automated test.

src/f32/funcs.rs
bitshifter: In addition to the request to make this an inline function and rename it, there are `mul_add` methods on `Vec3` and `Vec4` that could call this when `vec3sse2` and `vec4sse2` are enabled respectively.

src/f32/funcs.rs
```diff
@@ -139,6 +197,64 @@ pub(crate) mod sse2 {
...
+    pub(crate) unsafe fn m128_sin(v: __m128) -> __m128 {
+        // From DirectXMath: XMVectorSin
+        //
+        // 11-degree minimax approximation
+        //
+        // Note that XMVectorSin clamps the value here with XMVectorModAngles,
+        // but this adds about 25% overhead.
+        //
+        // // Force the value within the bounds of pi
+        // let mut x = m128_mod_angles(v);
```
bitshifter, line 218: Curious why this was commented out? From my experience with the approximation I was using here which I commented out, accuracy started to get a large error and angles grew beyond PI.

src/f32/funcs.rs
```diff
@@ -349,3 +465,25 @@ fn test_scalar_sin_cos() {
...
+
+#[test]
+#[cfg(vec4sse2)]
+fn test_sse2_m128_sin() {
+    use crate::Vec4;
+    use std::f32::consts::PI;
+
+    fn test_sse2_m128_sin_angle(a: f32) {
+        let v = Vec4::splat(a);
+        let v = unsafe { Vec4(sse2::m128_sin(v.0)) };
```
bitshifter, line 477: Does this function return the sin calculation in each lane? Can you pass in different angles in each lane in the input and get different results in each lane? Might be worth commenting one way or another. If it is actually returning a result in each lane then the test should test all lanes.

src/f32/funcs.rs
```diff
@@ -349,3 +465,25 @@ fn test_scalar_sin_cos() {
...
+        assert_approx_eq!(v.x(), a.sin(), 1e-6);
+    }
+
+    let mut a = -PI;
```
bitshifter, line 481: In `test_scalar_sin_cos` I'm testing a much larger range, and accuracy did turn out to be a problem for large input values. I think it would be better to copy that test, unless this sin is only being used by slerp and the input it expected to always be in the [-PI, PI] range then it might be OK to not include the `m128_mod_angles` call and not test a larger range. However it should be clearly commented that input needs to be [-PI, PI] and asserts added to check that is the case. Otherwise that `m128_mod_angles` call probably needs to got back in and this test should test across a large range of values. I guess if this is all sounding like too much work you could just use `f32:sin` and leave an optimised approximation for another day :)

aloucks:
`_mm_permute_ps` and `_mm_shuffle_ps` have `#[rustc_args_required_const(2)]` attribute set to enforce that the mask is constant. Using a macro is the only option.

```rust
pub(crate) unsafe fn m128_permute(a: __m128, mask: i32) -> __m128 {
    _mm_permute_ps(a, mask)
}
```

```
error: argument 2 is required to be a constant
   --> src\f32\funcs.rs:124:9
    |
124 |         _mm_permute_ps(a, mask)
    | 
```

aloucks:
@bitshifter I've added doc comments on `m128_sin` to explain that `sin` is calculated in each lane and that the error increases as the input drifts further from `[-PI, PI]`. 

The function is suitable for use in `slerp` as we know the input will always be within these bounds. I've also enabled the angle modulus so that it's now  implemented exactly as `XMVectorSin`. I compared this with the `C++` version of `XMVectorSin` and observed the same behavior.

I can remove the commented out test that attempts to test a larger range (and fails), if you'd like. I also checked using `f32::sin()` for each lane  and `slerp` was about twice as slow.

I have not updated vec3/vec4 to take advantage of the fused multiply add/sub. I started to do so, but I ran into an issue with importing the the macros and I didn't want to start re-organizing things in this PR.

src/f32/funcs.rs
```diff
@@ -139,6 +197,64 @@ pub(crate) mod sse2 {
...
+    /// the bounds of PI may produce an increasing error as the input angle
+    /// drifts from `[-PI, PI]`.
+    #[inline]
+    pub(crate) unsafe fn m128_sin(v: __m128) -> __m128 {
+        // From DirectXMath: XMVectorSin
+        //
+        // 11-degree minimax approximation
+        //
+        // Force the value within the bounds of pi
+        let mut x = m128_mod_angles(v);
```
bitshifter, line 219: Now that you are using mod angles again, is the accuracy OK with values outside of +/- PI? Just wondering if the function comment is still accurate?

aloucks: That depends on your definition of "OK". It seems that the error increases as you move further away from `[-PI, PI]` in either direction, but its still functions as a working `sin`. The output was rubbish outside the bounds of `PI` without `mod_angles`.

PR #65, opened 2020-06-28
Documentation clarifications

AlexEne:
I've added some clarification around Vec3/Vec3A change, as the docs.rs docs still mentions Vec3 almost everywhere and that was a bit confusing.


bitshifter:
Thanks for that.

repi:
@AlexEne nice!

PR #71, opened 2020-08-22
normalize_safe()

julhe:
I've added a ``normalize_safe()`` function to all vector structs. This should make life a little simpler when dealing with vectors that can be zero.

### Use Case
Say you an object that constantly seeks a target. To compute the direction, you would use:
``` rust
let direction = target - my_position;
let velocity = direction.normalize() * delta_time;
my_position += velocity;
```

However if ``target``  is equal to ``position``, velocity will be ``NaN``, which will also "infect" ``position``.
Fixing this would be:
```rust
let direction = target - my_position;
let velocity = if direction.length_squared() > f32::EPSILON {
   direction.normalize() * delta_time
} else {
   direction
};
my_position += velocity;
```
Which, to me, feels messy. Also, less expierenced users might use ``.length()``, which is slower or don't know about ``f32::EPSILON`` at all.

With ``normalize_safe()`` this would simply be:
``` rust
let direction = target - my_position;
let velocity = direction.normalize_safe() * delta_time;
my_position += velocity;
```

### DIFF
--- a/src/f32/vec2.rs
+++ b/src/f32/vec2.rs
@@ -176,6 +176,18 @@ impl Vec2 {
         self * self.length_reciprocal()
     }
 
+
+    /// Same as `normalize`, but will not normalize if `self` is zero length.
+    ///
+    /// Slightly slower than `normalize`.
+    #[inline]
+    pub fn normalize_safe(self) -> Self{
+        if self.length_squared() > f32::EPSILON {
+            return self.normalize();
+        }
+        return self;
+    }
+
     /// Returns the vertical minimum of `self` and `other`.
     ///
     /// In other words, this computes
--- a/src/f32/vec3.rs
+++ b/src/f32/vec3.rs
@@ -197,6 +197,17 @@ impl Vec3 {
         self * self.length_reciprocal()
     }
 
+    /// Same as `normalize`, but will not normalize if `self` is zero length.
+    ///
+    /// Slightly slower than `normalize`.
+    #[inline]
+    pub fn normalize_safe(self) -> Self{
+        if self.length_squared() > f32::EPSILON {
+            return self.normalize();
+        }
+        return self;
+    }
+
     /// Returns the vertical minimum of `self` and `other`.
     ///
     /// In other words, this computes
--- a/src/f32/vec3a.rs
+++ b/src/f32/vec3a.rs
@@ -511,6 +511,17 @@ impl Vec3A {
         }
     }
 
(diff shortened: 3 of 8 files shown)


bitshifter:
Hey, thanks for the PR.

I am not totally sure about this addition, as the tolerance used and what should be done in the almost zero length vector case could be a bit application specific.

I wouldn't want `glam` to assume that `f32::EPSILON` was appropriate in all cases. If a user was dealing with small numbers it would be possible to have a valid vector with a length less than `f32::EPSILON`. For example in Rust epsilon is `1.19209290e-07_f32` but the smallest possible float is `1.17549435e-38_f32` which is a considerably smaller number.

Looking at a few other engines:

* Unity returns zero vector, the tolerance is not documented which is a bit crap - https://docs.unity3d.com/ScriptReference/Vector3.Normalize.html
* Unreal takes a tolerance as an input parameter and returns zero vector if the squared magnitude is outside of the tolerance - https://docs.unrealengine.com/en-US/API/Runtime/Core/Math/FVector/GetSafeNormal/index.html
* DirectXMath just checks that length is greater than zero before dividing (so the resulting vector is zero if length is zero) - https://github.com/microsoft/DirectXMath/blob/master/Inc/DirectXMathVector.inl#L9828-L9845
* Defold doesn't check - https://defold.com/ref/stable/vmath/#vmath.normalize:v1
* Godot doesn't check - https://docs.godotengine.org/en/stable/classes/class_vector3.html#class-vector3-method-normalized
* This stack overflow post discusses a few other options, like flagging if the result is a degenerate case - https://stackoverflow.com/questions/722073/how-do-you-normalize-a-zero-vector

With `glam` I have generally taken the approach that it is up to the user to know to deal with degenerate cases.

So to not make any assumptions in `glam` perhaps one way of dealing with this would be (note I prefer checked to safe as checked/unchecked is already used for this kind of thing in std):

```
fn normalize_checked(self, tolerance: f32) -> Option<Vec3>;
```

But that ends up being reasonably verbose so might defeat the purpose of adding this as a convenience function.

Another option might be to specify what value should be used in the degenerate case:

```
fn normalize_checked(self, tolerance: f32, alternate: Vec3) -> Vec3;
```

If the DirectX approach of just checking length is greater than 0 to avoid divide by zero then the tolerance parameter could be dropped.

This might be better raised as an issue to see what other opinions are out there.

Note that in your code here, you should divide by the `sqrt` of `length_squared` to avoid calculating `length_squared` twice.

bitshifter:
Will close this for now and see where the discussion goes in #74

PR #75, opened 2020-08-30
Fix Mat4::transform_point3 to account for homogeneous w coordinate

termhn:
Currently, transform_point3 will fail on many perspective projections or other situations in which the homogeneous coordinate does not stay the same as it gets tossed out after the transformation. Instead, it must be renormalized to 1 by dividing by w.


bitshifter:
Ah whoops, that's a bit of an oversight on my part. Thanks for the PR.

PR #77, opened 2020-09-07
abs() to prevent inversion when point behind camera

aevyrie:
@termhn was helping me out when #75 was found. As a workaround, I manually implemented a transform_point3 with the same code as shown in that PR. 

However, I think there may be a bug with this. I recently had an issue where when vertices went behind the camera, `w` would negate and cause the vertices' coordinates to flip. The fix was to add an `.abs()` to the `let w` statement. See: https://github.com/aevyrie/bevy_mod_picking/issues/15#issuecomment-688431590

This is not my area of expertise, so I would appreciate feedback if this is incorrect. 🙂

### DIFF
--- a/src/f32/mat4.rs
+++ b/src/f32/mat4.rs
@@ -836,7 +836,7 @@ impl Mat4 {
         // res = self.w_axis.truncate() + res;
         // res
         let transformed = self.mul_vec4(other.extend(1.0));
-        let w_recip = transformed.w().recip();
+        let w_recip = transformed.w().abs().recip();
         Vec3::from(transformed.truncate() * w_recip)
     }
 


aevyrie:
CI error:
> warning: spurious network error (2 tries remaining): failed to get 200 response from `https://crates.io/api/v1/crates/crossbeam-epoch/0.8.2/download`, got 503
warning: spurious network error (1 tries remaining): failed to get 200 response from `https://crates.io/api/v1/crates/crossbeam-epoch/0.8.2/download`, got 503
error: failed to download from `https://crates.io/api/v1/crates/crossbeam-epoch/0.8.2/download`

> Caused by:
  failed to get 200 response from `https://crates.io/api/v1/crates/crossbeam-epoch/0.8.2/download`, got 503

bitshifter:
Hi @aevyrie . I'm not sure about this, I need to have a think about the math. Looking at similar functions in DirectXMath (https://github.com/microsoft/DirectXMath/blob/14bbb548f8720f293314133dd976148c482f3f17/Inc/DirectXMathVector.inl#L10685-L10701) they aren't taking the `abs` of `w` so based on that alone I suspect this isn't correct in the general case. I'll try look into it further though.

bitshifter:
In saying that the DirectXMath code is doing something different to what glam is doing, I'll look into it some more.

aevyrie:
I appreciate it! Let me know if there is anything I can do to help.

bitshifter:
I think the `abs` might be a red herring with regards to picking. I grabbed the latest master of `bevy_mod_picking` and changed the plane size to 50 and the camera centre point to (0, 0, 1) in the 3d_scene example and the picking flickers off and on for the plane when I move the mouse around.

aevyrie:
I never ended up uncovering the root cause of this bug, but I'm happy to defer to your judgement here.

Thinking about the consequences of applying abs() to `w`, this would mean that objects would scale up behind the camera (correct) but would not invert (incorrect). I think you are right in that this was a red herring, because the more I think about it, the vertices *should* invert if you are emulating the effects of projection.

PR #83, opened 2020-10-25
Implement Add<Quat> Sub<Quat>, Mul<f32> and Div<f32> for Quat

### LINKED ISSUE #82: impl Add<Quat> and Sub<Quat> for Quat
In short: Through a somewhat complicated chain of traits and supertraits in multiple different libraries, I've found myself needing glam::Quat to implement addition and subtraction. I've started working on a PR.

In long: I'm trying to implement skeletal animation for the bevy engine. They use glam, and rotations are represented by your quaternions. We're looking to use the splines crate to handle interpolation between keyframes, which is a very generic crate. It does however require the supplied type to implement addition and subtraction, and through a shim I've managed to get it working, but at the cost of user convenience. The splines crate provides implementations of its traits for a number of external math crates, but I am unable to add support for glam because of the requirement of these ops.

I understand that adding support for addition and subtraction ops might mislead users who are trying to combine rotations, but perhaps with good documentation we can mitigate this risk and steer them towards correct usage.

iwikal:
closes #82 

I tried to clarify in docs that these ops are not what you want to use most of the time. No editor I know of will show these docs when just using the `+` or `-` operator, but I couldn't think of a better way.

### DIFF
--- a/src/f32/quat.rs
+++ b/src/f32/quat.rs
@@ -455,22 +455,6 @@ impl Quat {
     }
 
     #[inline]
-    /// Adds two quaternions.
-    /// The sum not guaranteed to be normalized.
-    /// NB: Addition is not the same as combining the rotations represented by the two quaternions!
-    /// That corresponds to multiplication.
-    pub fn add_quat(self, other: Self) -> Self {
-        Self(self.0 + other.0)
-    }
-
-    #[inline]
-    /// Multiplies a quaternion with an f32.
-    /// The product is not guaranteed to be normalized.
-    pub fn mul_f32(self, other: f32) -> Self {
-        Self(self.0 * other)
-    }
-
-    #[inline]
     /// Multiplies two quaternions.
     /// If they each represent a rotation, the result will represent the combined rotation.
     /// Note that due to floating point rounding the result may not be perfectly normalized.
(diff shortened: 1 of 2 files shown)


src/f32/quat.rs
```diff
@@ -454,8 +454,25 @@ impl Quat {
         self.mul_vec3a(Vec3A::from(other)).into()
     }
 
+    #[inline]
+    /// Adds two quaternions.
+    /// The sum not guaranteed to be normalized.
+    /// NB: Addition is not the same as combining the rotations represented by the two quaternions!
+    /// That corresponds to multiplication.
+    pub fn add_quat(self, other: Self) -> Self {
```
bitshifter, line 462: I haven't been super consistent about adding functions as well as operators, but generally when I've added them it has been so there is a call by `&self` version, in particular for matrices which can be large. I think in this case it's not necessary and adding the implementation and comment on the operator itself would be sufficient.

src/f32/quat.rs
```diff
@@ -454,8 +454,25 @@ impl Quat {
...
+        Self(self.0 + other.0)
+    }
+
+    #[inline]
+    /// Multiplies a quaternion with an f32.
+    /// The product is not guaranteed to be normalized.
+    pub fn mul_f32(self, other: f32) -> Self {
+        Self(self.0 * other)
```
bitshifter, line 470: Same as above, the operator on its own is sufficient.

src/f32/quat.rs
```diff
@@ -454,8 +454,25 @@ impl Quat {
...
+    }
+
     #[inline]
     /// Multiplies two quaternions.
+    /// If they each represent a rotation, the result will represent the combined rotation.
```
bitshifter, line 475: I don't remember why I added this function, but regardless your comment update is fine.

bitshifter (changes requested):
Could you also please add unit tests for the operators you have added.

src/f32/quat.rs
iwikal: I added them mainly because I was following the example of mul_quat, and because I didn't think doc comments showed up in the generated docs. I was surprised to find they do, you just have to expand the trait impl to see them.

iwikal: Done!

src/f32/quat.rs
iwikal: Done!

bitshifter (approved):

iwikal:
Cool! Is there a plan for when the next release comes out? Is it a matter of months, weeks, days?

bitshifter:
Good question. I'm in the process of adding a new feature, which I've mostly with but still need to tidy up and document a bit. So probably a week or two.

Issue #5, opened 2019-07-11
no_std support


Lokathor:
You mention in the readme that you'd like to have no_std support, what are the current blockers?

bitshifter:
https://github.com/rust-lang/rfcs/issues/2505 is the blocker. I had a look at what nalgebra did, which was to let you link your own math functions. I'm not sure if that's the best approach (i.e. I really have no idea), I wasn't keen to just do what nalgebra did without knowing if it was actually a good solution. Potentially glam could just define it's own trig an sqrt functions. That would be an alternative approach to being no_std.

Lokathor:
oh, right. I've spent the past few days safe wrapping sse/sse2 for my core only crate and somehow i still totally forgot that there's no sqrt in core itself.

I'll look in to pure rust sqrt/sin/cos soon and if I find anything promising I'll bring it here too.

bitshifter:
Fixed by #88

PR #85, opened 2020-10-26
Initial SPIRV no_std implementation

repi:
This is a first early test implementation for our new `spriv-unknown-unknown` target that we have in our [Rust GPU compiler backend](https://shader.rs/) and to bring up `glam` to support it, this overall effort is tracked and discussed in https://github.com/EmbarkStudios/rust-gpu/issues/134

We currently require `no_std` so this enables `no_std` (related discussion: #5) when building glam for that target but then defines a small extension trait to bring in the `f32` functionality from `std`. This is right now defined directly in `glam` here in a separate `spirv.rs` and is missing some intrinsic like `tan` that we'll need we'll need to add to Rust core. 

Pretty happy that did get most things to compile this way, but some glam functionality has been stubbed out. Not too complex.

Expect this to live as a draft PR here for quite a while until we have settled on some of these intrinsics implementations, where they should live, and fixed some compiler backend issues with the this for now also. But having it as an open draft PR makes it more visible and easier to discuss approaches.


src/f32/spirv.rs
```diff
@@ -0,0 +1,88 @@
...
+    fn floor(self) -> Self;
+    fn ceil(self) -> Self;
+    fn exp(self) -> Self;
+    fn saturate(self) -> Self;
+
+    fn signum(self) -> Self;
+    fn copysign(self, sign: Self) -> Self;
+
+    // TODO: implement but need new intrinsic in core?
+    //fn tan(self) -> Self;
```
bjorn3, line 19: LLVM doesn't have a `tan` intrinsic, unlike all the other methods here.

src/lib.rs
```diff
@@ -1,3 +1,9 @@
+#![cfg_attr(target_arch = "spirv", no_std)]
```
khyperia, line 1: Shouldn't these be conditioned on the `std` feature, instead of `arch = spirv`? There's nothing spir-v specific here, and getting `no_std` working on non-spirv platforms seems useful to other folks.

repi: It could indeed be, though I did pick doing it on the arch here for now as this arch doesn't support std at all right now. But you are right if one wants to use no_std for other platforms (which is not supported now though), then one would use the `std` feature instead here. Could be cleaner also

bitshifter: It would be nice to add general `no_std` support although the mechanism might be different for spirv and the general case. For the general case I was thinking of trying https://github.com/rust-lang/libm. That might not be a good solution for your use case though? There is a long standing issue to support `no_std` in `glam`, I've just never go around to it - see https://github.com/bitshifter/glam-rs/issues/5.

build.rs
```diff
@@ -1,7 +1,8 @@
 use std::env;
 
 fn main() {
-    let force_scalar_math = env::var("CARGO_FEATURE_SCALAR_MATH").is_ok();
+    let force_scalar_math = env::var("CARGO_FEATURE_SCALAR_MATH").is_ok()
+        || env::var("CARGO_CFG_TARGET_ARCH").unwrap() == "spirv";
```
bitshifter, line 5: I would have though your target would get scalar math anyway as it lacks SSE2, or is that not the case? The other thing this does is turns off align16 on a bunch of types. Is that what you are wanting to achieve here?

src/f32/funcs.rs
```diff
@@ -1,3 +1,6 @@
+#[cfg(target_arch = "spirv")]
+use super::spirv::MathExt;
```
bitshifter, line 2: I guess you might eventually put this in its own crate so other libraries can use it?

src/f32/spirv.rs
bitshifter: There's a `tanf` implementation here if you need one https://github.com/rust-lang/libm/blob/master/src/math/tanf.rs.

src/f32/funcs.rs
repi: yes we do have it in our `spirv-std` also: https://github.com/EmbarkStudios/rust-gpu/blob/a4a34243c0589aadf7496346a8c871d4c2f30c10/spirv-std/src/math_ext.rs

But it is not published yet but leaning towards simply publishing that and using it here

build.rs
repi: the SSE2 support did get enabled when I didn't have this, maybe it is still somehow in the `CARGO_CFG_TARGET_FEATURE`, haven't debugged it more carefully so just did a quick solution here.

Was intending of not having any alignment and using "scalar" types

src/lib.rs
repi: would indeed be nice to have general no_std support, and depending on approach, if one does use core intrinsics like here + libm there is a potential of using the same implementation. 

because we talked a bit about using `libm` also and in the backend replace some of the functions with SPIRV intrinsics when available as a potential solution for implementing the SPIRV intrinsics that Rust core or LLVM does not expose.

but we'll see how this evolves, definitely surprisingly similar problems here on CPU and GPU no_std :)

bitshifter:
I'm not sure it's of use to your SPIRV efforts but I fixed `no_std` in this branch https://github.com/bitshifter/glam-rs/tree/no_std. Needs a little tidying up, but I might merge it to master soon.

It requires `--no-default-features --features libm`

repi:
@bitshifter thanks will check out!

src/lib.rs
```diff
@@ -166,6 +166,25 @@ The minimum supported version of Rust for `glam` is `1.36.0`.
 */
 #![doc(html_root_url = "https://docs.rs/glam/0.10.2")]
 #![cfg_attr(not(feature = "std"), no_std)]
+#![cfg_attr(target_arch = "spirv", feature(register_attr, repr_simd))]
+
+#[cfg(all(target_arch = "spirv", feature = "std"))]
+compile_error!("`std` feature is not supported when building for SPIRV");
+
+#[cfg(all(target_arch = "spirv", feature = "glam-assert"))]
```
bitshifter, line 174: I noticed in the Cargo.toml you had disabled the optional `bytemuck` dependency for `spirv`. Is it worth doing that for the rest?

bitshifter (commented):
What do you mean I can't printf debug on the GPU :)

build.rs
khyperia: Tried debugging this to figure out what's going on here, and could repro it _once_, but then once I `cargo clean`'d and rebuilt with a bunch of debugging `println!` whatnots, I... could no longer repro it, no matter what I tried.

I'm guessing it may be a build cache bug, possibly due to lack of [cargo:rerun-if-env-changed=NAME](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-env-changed) on these variables, but that's a wild guess. In any case, I'm going to remove this as it seems to work fine without it.

khyperia: Right, managed to repro it now, the actual problem is that we're building `rust-gpu` crates in a nested cargo process. Cargo, helpfully, if there are no target features, does not reset-to-empty/unset `CARGO_CFG_TARGET_FEATURE`. So that means the outer cargo process's `CARGO_CFG_TARGET_FEATURE` "leaks" into the inner process, causing `CARGO_CFG_TARGET_ARCH` to say `spirv` but `CARGO_CFG_TARGET_FEATURE` has x86_64 features. Ugh.

This is our own problem to fix, though, not glam's, so not having this bit of code in glam is the right way to go.

src/lib.rs
khyperia: Ended up just removing this cfg out, since it didn't seem to be doing anything (was needed in an old draft, but not anymore)

In any case, I think this PR is now ready for review and merge consideration! :tada:

bitshifter (approved):
LGTM

repi:
Wooho! 🎉🚀

Issue #15, opened 2019-07-23
Swizzle support


bitshifter:
It would be good to support generic swizzles.

Ideally this would support swizzling to return the same size vector but a long term goal would be to provide swizzling into a vector of a different size, smaller or larger.

The problem with SSE2 shuffles is the mask needs to be constant. The intrinsic itself is not a `const` function so I can't provide a method that will take a mask or at least I don't think that's possible. There are a couple of alternatives: 

- Generate shuffle methods for all possible permutations. This is what pathfinder does but that's 255 methods for swizzling a `Vec4` to another `Vec4`, potentially more if swizzling to a different size is supported. My main concern with this approach is making sure it doesn't make a mess of the docs.
- Use macros - the underlying `__m128` is accessible so I think a macro solution would be possible. I would probably provide `vec4_swizzle`, `vec3_swizzle`, and `vec2_swizzle` so that the type is known and the number of lanes specified in the macro could be used to specify the output size.

Lokathor:
Just write a program to auto-generate all the swizzles

it's only [4000 lines of code give or take](https://github.com/Lokathor/hektor/blob/master/src/swizzle4.rs) once the program prints it out ;3

bitshifter:
I wrote a program to generate all the swizzles :)

PR #86, opened 2020-10-27
Add vector swizzle functions.

### LINKED ISSUE #14: Make `dup_x` etc. methods public
These are currently `pub(crate)` but they look like they'd be useful externally based on https://github.com/cessen/psychopath/commit/88e7365bc45b737b4fb0d01943271e51dceca26a#diff-7e162a40d8f4de03dc9ab929005466cfR65

bitshifter:
Fixes #14 and #15.


cessen:
I am *so* excited about this!  Great work!

aloucks:
I kind of liked directx's nomenclature, e.g:  `splat_x` when it's replicating a single element to all lanes. It has the added benefit of having the same function name regardless of whether or it's a `Vec4` vs `Vec3`, for example.

bitshifter:
@aloucks I did used to have `dup_x` methods on vectors but they were never `pub` and I've replaced usage with swizzles. The `splat_x` use case is pretty common though, so could potentially add them back. Would mean there's a couple way of achieving the same thing but you wouldn't need to include the swizzle traits to use them. Will have a think about it.

PR #88, opened 2020-10-30
Add no_std support, requires libm.

bitshifter:


khyperia:
How are you supposed to use this? I tried creating a new empty test crate, adding this line to `Cargo.toml`, and there's compile errors. (Also fails with `version = "0.10.0"` instead of `git = "..."`)

```
[dependencies]
glam = { git = "https://github.com/bitshifter/glam-rs.git", default-features = false, features = ["libm"] }
```

Lots of errors, but one example:

```
error[E0599]: no method named `signum` found for type `f32` in the current scope
   --> /home/khyperia/.cargo/git/checkouts/glam-rs-bbec0e101ff938b0/454e9c8/src/f32/vec3.rs:422:23
    |
422 |             y: self.y.signum(),
    |                       ^^^^^^ method not found in `f32`
    |
    = help: items from traits can only be used if the trait is in scope
    = note: the following traits are implemented but not in scope; perhaps add a `use` for one of them:
            candidate #1: `use crate::num_traits::float::FloatCore;`
            candidate #2: `use crate::num_traits::Float;`
            candidate #3: `use crate::num_traits::real::Real;`
            candidate #4: `use crate::num_traits::Signed;`
```

---

I think the issue is that even though there's this:

```
#[cfg(all(not(feature = "std"), feature = "libm"))]
extern crate num_traits;
```

there still needs to be a `use` for the traits inside `num_trait`, which I don't see anywhere.

PR #90, opened 2020-10-31
Added distance and distance_square to calculate distance between two points

### LINKED ISSUE #72: Distance between two points
It would be nice to implement two functions `distance` and `distance_squared` which returns the Euclidean distance between two points. In other words, `distance` would return `sqrt((x1-x2)^2 + (y1-y2)^2 + (z1-z2)^2)` while `distance_squared` would return `(x1-x2)^2 + (y1-y2)^2 + (z1-z2)^2`.

Neo-Zhixing:
These two operations should be common enough to merit their own methods.

Closes #72 

This implementation reused the `length` and `length_squared` functions so they automatically have SIMD.

### DIFF
--- a/src/f32/vec2.rs
+++ b/src/f32/vec2.rs
@@ -199,6 +199,14 @@ impl Vec2 {
         self.length().recip()
     }
 
+    /// Computes the Euclidean distance between two points.
+    #[inline]
+    pub fn distance(self, other: Vec2) -> f32 { (self - other).length() }
+
+    /// Compute the squared Euclidean distance between two points.
+    #[inline]
+    pub fn distance_squared(self, other: Vec2) -> f32 { (self - other).length_squared() }
+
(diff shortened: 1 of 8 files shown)


src/f32/vec2.rs
```diff
@@ -199,6 +199,14 @@ impl Vec2 {
         self.length().recip()
     }
 
+    /// Computes the Euclidean distance between two points.
+    #[inline]
+    pub fn distance(self, other: Vec2) -> f32 { (self - other).length() }
```
aloucks, line 204: Why are you putting the function bodies all on one line?

Neo-Zhixing: Just used cargo fmt to format all the codes.

Thanks for the heads up!

bitshifter:
Thanks!

PR #93, opened 2020-11-09
Implement arithmetic traits

msvbg:
This PR implements the arithmetic traits `Sum` and `Product`, allowing you to do this stuff:
```rust
let vecs: Vec<Vec2> = vec![Vec2::new(1.0, 2.0), Vec2::new(3.0, 4.0)];

assert_eq!(v.iter().sum::<Vec2>(), Vec2::new(4.0, 6.0));
```

Apologies for the PR spam, I messed up the last PR.

### DIFF
--- a/src/f32/mat2.rs
+++ b/src/f32/mat2.rs
@@ -8,6 +8,9 @@ use core::{
     ops::{Add, Mul, Sub},
 };
 
+#[cfg(feature = "std")]
+use std::iter::{Product, Sum};
+
 const ZERO: Mat2 = const_mat2!([0.0; 4]);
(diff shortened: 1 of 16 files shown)


bitshifter:
Hi there, thanks for the PR.

I always like to understand the use case for new glam features. Can you elaborate how you are using this feature? I understand the code example, I'm more asking what specifically you are doing that this feature helps with.

With Quaternions, does it make sense to sum them?

msvbg:
I implemented this for vectors first. I was implementing an algorithm which applies forces at vertices of a grid. I thought it was clean to be able to use `.sum()` calculate the resultant force at every given point.

I then added implementations for matrices and quaternions for completeness, not for any particular use case. Summing quaternions is unlikely to be very useful.

PR #96, opened 2020-11-15
Add perp() method to Vec2

basro:
A perp() method for Vec2 that returns (-y, x) is pretty useful.
Given that there's already a perp_dot I think it makes sense to have this too.


bitshifter:
Thanks!

Issue #74, opened 2020-08-29
Behaviour for normalization of zero-vectors


julhe:
Hey all! In #71 I proposed to add ``normalized_safe()`` as a quick way to deal with vectors that might be zero (see the example). However @bitshifter said that this should be subject of discussion (on which I agree). 

So the questions are:

-  Do we need to simplify normalization of zero vectors?
-  If so, how should that function look like?

bitshifter:
Thanks for opening the issue @julhe.

Functions that might benefit from having checked versions:
* `Vec2`, `Vec3`, `Vec4` and `Quat` `normalize`
* `Mat2`, `Mat3`, `Mat4` `inverse`

There are other functions that may depend on valid inputs, these ones in particular the input can be variable and they will divide by zero if the input is invalid.

Some math libraries always check, some never check and some provide separate methods. A quick survey found:
* Unity returns zero vector, the tolerance is not documented which is a bit crap - https://docs.unity3d.com/ScriptReference/Vector3.Normalize.html
* Unreal takes a tolerance as an input parameter and returns zero vector if the squared magnitude is outside of the tolerance - https://docs.unrealengine.com/en-US/API/Runtime/Core/Math/FVector/GetSafeNormal/index.html
* DirectXMath just checks that length is greater than zero before dividing (so the resulting vector is zero if length is zero) - https://github.com/microsoft/DirectXMath/blob/master/Inc/DirectXMathVector.inl#L9828-L9845
* Defold doesn't check - https://defold.com/ref/stable/vmath/#vmath.normalize:v1
* Godot doesn't check - https://docs.godotengine.org/en/stable/classes/class_vector3.html#class-vector3-method-normalized

This stack overflow post discusses a few other options, like flagging if the result is a degenerate case - https://stackoverflow.com/questions/722073/how-do-you-normalize-a-zero-vector

If providing checked functions there are a couple of decisions that need to be made:
* what epsilon value should be used for length? If just trying to avoid divide by zero then checking length is > 0.0 would be sufficient, but users may be interested in catching small values. If the epsilon value is greater than zero I believe it should be an input parameter
* what to do if the input is degenerate? My feeling here was a closure might offer the best flexibility for users to decide what is the appropriate behavior in this case. Some libraries return the input, others return zero and others just perform the divide by zero and return NaNs.

For example:

```
// if length is zero then F is called
fn normalize_or_else<F>(self, f: F) -> Self;
```

[...]

PR #98, opened 2020-11-18
Add `fn is_finite(&self) -> bool` member to all types

emilk:
Having a function like this is super-useful to guard against non-finite values. For instance:

``` rust
let v = foo.cross(bar) / det;
let v = v.normalize():
if v.is_finite() {
   Some(v)
} else {
   None
}
```

or `debug_assert!(transform.is_finite());`

## Naming

There is already a `is_nan` which returns a mask. This sets a precedence that something called `is_finite` should also return a mask. However, `is_entirely_finite` doesn't feel like a great name to me.
Maybe functions that return a mask should have a `m` suffix or something instead?

Anyway, I'll be happy to change the name of this, this is just a first draft!

### DIFF
--- a/src/f32/mat2.rs
+++ b/src/f32/mat2.rs
@@ -150,6 +150,13 @@ impl Mat2 {
     //     }
     // }
 
+    /// Returns `true` if, and only if, all elements are finite.
+    /// If any element is either `NaN`, positive or negative infinity, this will return `false`.
+    #[inline]
+    pub fn is_finite(&self) -> bool {
+        self.x_axis.is_finite() && self.y_axis.is_finite()
+    }
+
     /// Returns the transpose of `self`.
     #[inline]
     pub fn transpose(&self) -> Self {
--- a/src/f32/mat3.rs
+++ b/src/f32/mat3.rs
@@ -255,6 +255,13 @@ impl Mat3 {
     //     }
     // }
 
+    /// Returns `true` if, and only if, all elements are finite.
+    /// If any element is either `NaN`, positive or negative infinity, this will return `false`.
+    #[inline]
+    pub fn is_finite(&self) -> bool {
+        self.x_axis.is_finite() && self.y_axis.is_finite() && self.z_axis.is_finite()
+    }
+
     /// Returns the transpose of `self`.
     #[inline]
     pub fn transpose(&self) -> Self {
--- a/src/f32/mat4.rs
+++ b/src/f32/mat4.rs
@@ -344,6 +344,16 @@ impl Mat4 {
     //     }
     // }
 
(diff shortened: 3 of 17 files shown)


bitshifter:
Hey there,

Returning a mask is optimal for types that are using SIMD, but not for types that aren't since they lose short circuiting. So it has been in the back of my mind that for common things I might want also have methods that return a `bool` instead of or in addition to returning a mask. This like finite and NaN checks would seem like good candidates. Naming wise I would be tempted to go with the prefix `all` and `any` because they match the mask methods and it is hopefully less ambiguous than using `is` on a vector. I could imagine a use case for `any_nan` and `all_fininte` but I don't think `all_nan` or `any_fininte` would be necessary right now. 

If that sounds good to you, I would suggest renaming to `all_finite`. I  think I will rename `is_nan` to `any_nan` and make it return a `bool`. If someone wanted a `NaN` mask for some reason I could add `cmp_nan` later, which would match the other methods that return a mask.

bitshifter:
Also, for the specific example of performing a checked normalize you may be interested in #74.

emilk:
`all_finite` sounds good to me!

bitshifter:
Yeah that is a good point about it being different for the type of `.x` and I do try to match the method names used on the primitive Rust types where possible. 

The main issue I have with `is_finite` and `is_nan` on a vector is internally I think it would make the most sense for one to be an `and` of `is_finite` for all elements and the other to be an `or` of `is_nan` for all elements. To me `is` suggests `all` so `is_nan` might be a bit ambiguous, on the outside to me it sounds like it would only be true if all elements were `NaN`, but usually I think people want to know if any element is `NaN`.

I would plan to rename `is_nan -> mask` to `any_nan -> bool` be consistent with this PR.

bitshifter:
I guess I could keep `is_finite` and just rename `is_nan` to `any_nan` or `contains_nan`. That would also be clear I think.

bitshifter:
Perhaps leave the name as is. It would be good to add tests though. At some point I'll do a SIMD implementation so it will save me some time.

bitshifter:
Now that I've merged it and I was looking at renaming `is_nan` to match, I've found that I do use `is_nan` as a mask for the SIMD implementation of `signum`. Which I think it bringing me back around to my original naming thought which would be to make `is` operations on vectors return a mask / vector of bools and to have `any` or `all` methods return a single bool.

bitshifter:
Naming is hard :)

emilk:
Indeed :)

I'm still for naming consistency with `std::f32` and friends, but then again I am not using masks right now 🤷

bitshifter:
In the end I've gone with `is_nan -> bool`, `is_finite -> bool` and `is_nan_mask -> mask`.

PR #101, opened 2020-11-30
Methods to clamp vector length

### LINKED ISSUE #100: Would clamp methods be welcome?
A common thing I do in my game projects is clamp vector length to a certain max, min, or both.  Would functionality like that be welcome in `glam`? I am willing to open a PR if this would be useful for glam itself (though I'd likely need assistance with the SIMD part).  I'm thinking something along the lines of (for Vec2, for example):

```rust
    /// Returns a `Vec2` with a length no less than `min` and no more than `max`
    pub fn clamp(&self, min: f32, max: f32) -> Self {
        if self.length() < min {
            self.normalize() * min
        } else if self.length() > max {
            self.normalize() * max
        } else {
            *self
        }
    }

    /// Returns a `Vec2` with a length no more than `max`
    pub fn clamp_max(&self, max: f32) -> Self {
        if self.length() > max {
            self.normalize() * max
        } else {
            *self
        }
    }

    /// Returns a `Vec2` with a length no less than `min`
    pub fn clamp_min(&self, min: f32) -> Self {
        if self.length() < min {
            self.normalize() * min
        } else {
            *self
        }
    }
```

### Additional References
- [`ClampVectorSize` Blueprint Node in Unreal Engine 4](https://docs.unrealengine.com/en-US/BlueprintAPI/Math/Vector/ClampVectorSize/index.html)
- [`Vector3.ClampMagnitude` in Unity](https://docs.unity3d.com/ScriptReference/Vector3.ClampMagnitude.html)
- [`Vector2 clamped` in Godot](https://docs.godotengine.org/en/stable/classes/class_vector2.html#class-vector2-method-clamped)

CleanCut:
An implementation of `clamp_length*` methods for `Vec2`, `Vec3`, `Vec3A`, and `Vec4`.

This PR is based off of `0.11.0`, but I can see that there is a pretty heavy refactor going on in the `refactor` branch to produce the `Vec*` implementations through macros.  I'm willing to reimplement this PR off of `refactor` if that's helpful. (probably after it's in a stable state?)

Resolves #100

### DIFF
--- a/src/f32/vec2.rs
+++ b/src/f32/vec2.rs
@@ -401,6 +401,36 @@ impl Vec2 {
             angle
         }
     }
+
+    /// Returns a `Vec2` with a length no less than `min` and no more than `max`
+    pub fn clamp_length(&self, min: f32, max: f32) -> Self {
+        let length_sq = self.length_squared();
+        if length_sq < min * min {
+            self.normalize() * min
+        } else if length_sq > max * max {
+            self.normalize() * max
+        } else {
+            *self
+        }
+    }
+
+    /// Returns a `Vec2` with a length no more than `max`
+    pub fn clamp_length_max(&self, max: f32) -> Self {
+        if self.length_squared() > max * max {
+            self.normalize() * max
+        } else {
+            *self
+        }
+    }
+
+    /// Returns a `Vec2` with a length no less than `min`
+    pub fn clamp_length_min(&self, min: f32) -> Self {
+        if self.length_squared() < min * min {
+            self.normalize() * min
+        } else {
+            *self
+        }
+    }
 }
 
 impl fmt::Display for Vec2 {
--- a/src/f32/vec2.rs
+++ b/src/f32/vec2.rs
@@ -406,9 +406,9 @@ impl Vec2 {
     pub fn clamp_length(&self, min: f32, max: f32) -> Self {
         let length_sq = self.length_squared();
         if length_sq < min * min {
-            self.normalize() * min
+            *self * (length_sq.sqrt().recip() * min)
         } else if length_sq > max * max {
-            self.normalize() * max
+            *self * (length_sq.sqrt().recip() * max)
         } else {
             *self
         }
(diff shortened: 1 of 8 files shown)


bitshifter:
Hi! Yeah there is a big refactor going on. I might be working on that for some time. It's probably easier if I apply changes to that branch myself since it's still a work in progress.

src/f32/vec2.rs
```diff
@@ -401,6 +401,36 @@ impl Vec2 {
             angle
         }
     }
+
+    /// Returns a `Vec2` with a length no less than `min` and no more than `max`
+    pub fn clamp_length(&self, min: f32, max: f32) -> Self {
+        let length_sq = self.length_squared();
+        if length_sq < min * min {
+            self.normalize() * min
```
bitshifter, line 409: normalize will recalculate length squared again so it would be better to reuse the `length_sq` value that's already been calculated, e.g. replace normalize with something like `self * (length_sq.sqrt().recip() * min)`.

The normalize function is doing `self * self.length_squared().sqrt().recip()`

bitshifter (changes requested):

src/f32/vec2.rs
bitshifter: Realized I missed the recip when I mentioned this on the issue.

CleanCut: Ah, so that's what was going on with your example values. All I knew was the value was way off. I'll update this PR tomorrow, so at least it's a valid reference for the eventual implementation.

CleanCut: Apologies, real life intervened so I wasn't able to turn this around when I had intended.

I've just pushed the optimization updates. 👍🏼  How is the macro-ification coming? Is it at a good state where I should look at porting this to a macro?

bitshifter:
@CleanCut I've merged the big refactor back to master. If you want to update this PR, you can add the methods once to the `impl_vecn_float_methods` macro in `src/vec.rs`. Each test will have a macro for implementing float methods as well, e.g. `impl_vec2_float_tests`. These methods and tests will be added for for `f32` types and `f64` types.

CleanCut:
Closing in favor of #121

PR #103, opened 2020-12-03
Add From<(VecN-1, f32)> for VecN

XAMPPRocky:
This PR adds some convenience `From` impls that allow you concat a tuple of a vector with a f32 into the next biggest vector similar to GLSL's initialisers such as `vec4(vec3, w)` without having to write out every coordinate.

```rust
let v2 = Vec2::from((1.0, 2.0));
let v3 = Vec3::from((v2, 3.0));
let v4 = Vec4::from((v3, 4.0));
```

### DIFF
--- a/src/f32/vec3.rs
+++ b/src/f32/vec3.rs
@@ -756,3 +756,13 @@ impl<'a> Product<&'a Self> for Vec3 {
         iter.fold(ONE, |a, &b| Self::mul(a, b))
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::{Vec2, vec3};
+    
+    #[test]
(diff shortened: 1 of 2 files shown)


bitshifter (changes requested):
Seems like a good solution, I waver between using From and more explicit functions sometimes because I think explicit functions are a bit more discoverable. This seems fine though.

Happy to merge this but I'd like tests for these new methods..

XAMPPRocky:
@bitshifter Updated with tests.

bitshifter:
Thanks!

Boscop:
Could you please also add `From<{Vec2, Vec2)> for Vec4`? :)

PR #106, opened 2020-12-04
Make Vec3 be #[repr(spirv)] on SPIR-V target

termhn:
I think this is a good idea but we should have @khyperia, @eddyb, @Jasper-Bekkers chime in. Without this, Vec3 is basically useless on spirv targets which doesn't feel great. And just using Vec3A everywhere instead has several pain points, like not being the default result of swizzles and Matrix4::transform_{point|vec}3 not working without conversions.

### DIFF
--- a/src/macros.rs
+++ b/src/macros.rs
@@ -62,17 +62,54 @@ macro_rules! const_mat2 {
 /// ```
 #[macro_export]
 macro_rules! const_mat3 {
-    ($f32x9:expr) => {
-        unsafe { $crate::f32::F32x9Cast { f32x9: $f32x9 }.mat3 }
-    };
-    ($col0:expr, $col1:expr, $col2:expr) => {
+    ($f32x9:expr) => {{
+        #[cfg(not(target_arch = "spirv"))]
+        unsafe {
+            $crate::f32::F32x9Cast { f32x9: $f32x9 }.mat3
+        }
+
+        // special path for SPIRV, without this we get:
+        //
+        //    error[E0080]: it is undefined behavior to use this value
+        //    --> glam-rs\src\f32\mat3.rs:6:1
+        //     |
+        //   6 | const ZERO: Mat3 = const_mat3!([0.0; 9]);
+        //     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ type validation failed: encountered uninitialized bytes at .z_axis.1, but expected initialized plain (non-pointer) bytes
+        //     |
+        //     = note: The rules on what exactly is undefined behavior aren't clear, so this check might be overzealous. Please open an issue on the rustc repository if you believe it should not be considered undefined behavior.
+        #[cfg(target_arch = "spirv")]
+        $crate::Mat3 {
+            x_axis: Vec3 { x: $f32x9[0], y: $f32x9[1], z: $f32x9[2] },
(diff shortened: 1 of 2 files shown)


khyperia:
While the `const_mat3` change fixes it, there's probably a cleaner fix. I think the error is because the alignment of `Vec3` changes when it's made to be `repr(simd)`, going from 12 bytes to 16 bytes, and so the cast from a `[f32; 9]` is no longer valid - there needs to be padding bytes to make it work. In any case, might be nice to comment that that's the reason why there's a special path, it's not because rust-gpu is weird and special, but rather because `repr(simd)` changes the alignment and so needs special handling (and would happen on the CPU as well).

Jasper-Bekkers:
I think I'd be on board with this change, but it's slightly confusing that there will now be a mismatch in size between CPU and GPU types, the biggest problem I anticipate with this would be constant buffer uploads for example or any other code where data gets shared between CPU and GPU.

bitshifter:
I am fine with this if it makes sense for spriv. Would there ever be any need for a non simd and thus not 16 byte aligned `Vec3` though?

src/macros.rs
```diff
@@ -62,17 +62,54 @@ macro_rules! const_mat2 {
...
+        //
+        //    error[E0080]: it is undefined behavior to use this value
+        //    --> glam-rs\src\f32\mat3.rs:6:1
+        //     |
+        //   6 | const ZERO: Mat3 = const_mat3!([0.0; 9]);
+        //     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ type validation failed: encountered uninitialized bytes at .z_axis.1, but expected initialized plain (non-pointer) bytes
+        //     |
+        //     = note: The rules on what exactly is undefined behavior aren't clear, so this check might be overzealous. Please open an issue on the rustc repository if you believe it should not be considered undefined behavior.
+        #[cfg(target_arch = "spirv")]
+        $crate::Mat3 {
```
bitshifter, line 81: This should work for all platforms shouldn't it? I wrote these macros before the members were public so it wouldn't have been possible to do it this way. It might be cleaner to do:

```
/// Creates a `Mat3` from three column vectors that can be used to initialize a constant value.
///
/// ```
/// # #[macro_use] extern crate glam;
/// use glam::{const_mat3, Mat3};
/// const ZERO: Mat3 = const_mat3!([0.0; 9]);
/// const IDENTITY: Mat3 = const_mat3!([1.0, 0.0, 0.0], [0.0, 1.0, 0.0], [0.0, 0.0, 1.0]);
/// ```
#[macro_export]
macro_rules! const_mat3 {
    ($f32x9:expr) => {
        Mat3 {
            x_axis: const_vec3!([$f32x9[0], $f32x9[1], $f32x9[2]]),
            y_axis: const_vec3!([$f32x9[3], $f32x9[4], $f32x9[5]]),
            z_axis: const_vec3!([$f32x9[6], $f32x9[7], $f32x9[8]]),
        }
    };
    ($col0:expr, $col1:expr, $col2:expr) => {
        Mat3 {
            x_axis: const_vec3!($col0),
            y_axis: const_vec3!($col1),
            z_axis: const_vec3!($col2),
        }
    };
}
```

Either way I think it should be possible to use the same technique for spriv and non-spriv.

bitshifter (dismissed):

PR #116, opened 2020-12-29
Move CI from Travis to GitHub Actions

### LINKED ISSUE #84: Migrate from Travis to GitHub Actions
Having a github action for cargo deny seems to have broken travis builds for PRs, which is unfortunate. I think most likely the simplest thing to do is to migrate entirely to github actions. Also needs to work with code coverage.

Edit: the travis builds still happen for PRs but they don't appear in GitHub, at least they don't on success.

XAMPPRocky:
resolves #84


bitshifter:
Awesome, thanks for this!

PR #117, opened 2021-01-12
Add support for `f64`, `i32`, `u32` primitive types.

### LINKED ISSUE #70: Support f64 and i32
Once the glam API is stable (e.g. 1.0) I think it would make sense to add support for f64 and i32. I still would avoid using generics, since I couldn't easily provide SIMD implementation without specialisation, and even if specialisation was available it would complicate the API using generics. So perhaps some macros behind the scenes or just copy and paste. A lot of functionality is pretty simple so I don't think duplication is the end of the world, once the API is stable.

### LINKED ISSUE #99: Internal rewrite
I'm opening this for visibility, particularly for #85 since internal changes will require rework for any one with active glam forks.

The primary design goals of glam have always been having a simple user facing API and trying to get the best performance for scalar math by using SIMD under the hood where possible. Part of keeping things simple was limiting glam to only supporting `f32`, since the majority of the time I've only needed `f32`. Inside glam I have also opted to keep things simple and accepted to a bit of code duplication to avoid complicating the implementation with generics, traits and macros.

It has been a pleasant surprise that other people have liked and adopted the crate. Other people, it turns out, sometimes want different things from me. For example `f64` or `i32` support. Or support for different architectures like WASM or Neon. Long term I think supporting these things is going to tip the scales away from code duplication in favour of a more complex implementation to reduce the development and maintenance burden on myself and contributors. I do not wish to burden users with additional complexity so the public API should remain the same.

This refactor should help with:
* #70 
* #95 
* Adding support for other SIMD architectures (although this could also be achieved by using the `packed_simd` or `wide` crates.
* Possible `Vec2` optimisations - that is internally loading into SIMD, performing the operation and storing back to the Vec2 (see `pathfinder_geometry`)
* Possibly exposing a low level API similar to DirectX math that can be used to write high level APIs that are different to glam

At a high level the idea is to add storage types and implement operations on those storage types. Similar to how DirectXMath has its `XMVECTOR` type and functions that perform Vector3, Vector4, Quaternion operations on the one type. It's not a friendly API for users but it makes sense internally. My idea is the make traits for these methods and implement them for the different storage types that I want to support, e.g. `struct XYZ<T> { x: T, y: T, z: T}` or `__m128` could both implement a `Vector3` trait. Using traits does make it slightly easier to support `f32`, `f64`, `i32` and so on without having to add duplicate implementations. The high level types like `Vec3`, or `DVec3` can be defined using a single macro that specifies the scalar type.

I have a work in progress branch here https://github.com/bitshifter/glam-rs/tree/refactor.

Things to do:
* [x] `f64` types - `DVec2`, `DVec3`, `DVec4`, `DQuat`, `DMat2`, `DMat3`, `DMat4`
* [x] `i32` types - `IVec2`, `IVec3`, `IVec4`
* [x] `u32` types - `UVec2`, `UVec3`, `UVec4`
* [x] `bool` types - `BVec2`, `BVec3`, `BVec4` - mask type can just be used when SIMD is available
* [x] Check performance
* [x] Check documentation makes sense
  * [x] Document internal architecture
  * [x] Improve module level module docs (maybe split types by module)
  * [x] Document Deref usage
  * [x] Document `as` cast methods
* [x]  Check debugging experience (VS Code)
  * [x] Using `map` functions obsfucates implementation a bit
  * [x] Need to check matrix debugging.
* [x] Tests for const macros
* [x] Swizzles for all types
* [x] Casts between different types
* [x] Support optional features (serde, mint, rand) for the new types
* [x] Remove build.rs
* [x] Check with rust-gpu
* [x] Check with bevy
* [x] Check with macroquad (can't - it's using an older version of glam)
* [x] Check build times
  * [x] Make `num-traits` optional to reduce build times

bitshifter:
* Added `f64` primitive type support
  * vectors: `DVec2`, `DVec3` and `DVec4`
  * square matrices: `DMat2`, `DMat3` and `DMat4`
  * a quaternion type: `DQuat`
* Added `i32` primitive type support
  * vectors: `IVec2`, `IVec3` and `IVec4`
* Added `u32` primitive type support
  * vectors: `UVec2`, `UVec3` and `UVec4`
* Added `bool` primitive type support
  * vectors: `BVec2`, `BVec3`, `BVec3A`, `BVec4` and `BVec4A`

Internally glam has been significantly refactored to support 4 different primitive type implementations as well as supporting SSE2 for many types.

This is a breaking change primarily due to `Vec2Mask`, `Vec3Mask` and `Vec4Mask` being removed and being replaced by `BVec2`, `BVec3`, `BVec3A`, `BVec4` and `BVec4A`.

Fixes #70.
Fixes #99.


Cargo.toml
```diff
@@ -40,7 +39,6 @@ libm = ["num-traits/libm"]
 [dependencies]
 bytemuck = { version = "1.4", optional = true, default-features = false }
 mint = { version = "0.5", optional = true, default-features = false }
-# enabled by the libm feature, required when building no_std
```
bitshifter, line 43: Add this comment back

README.md
```diff
@@ -105,7 +115,6 @@ The design of this library is guided by a desire for simplicity and good
 performance.
 
 * No traits or generics for simplicity of implementation and usage
```
bitshifter, line 117: Should be more precise here - no traits or generics in the public interface

PR #120, opened 2021-01-15
Add Vec2::X Vec4::W etc as shorter versions of unit_x() et al

emilk:
An ergonomic win:

Before: `pos + 5.0 * Vec3::unit_y()`
After: `pos + 5.0 * Vec3::Y`

This is very similar to how one in math uses `Y` as the name of the basis vector, so that `5 * Y` means "A vector in the Y direction of length 5".

Sorry for not creating an issue and jumping straight to a PR - it was such a simple thing to add so I thought it would just be easier to implement and then discuss it in the PR.

If this is appreciated we could also consider adding `Vec3::ZERO` and `Vec3::ONE`, though it is less of a win since it is not that much shorter. We could also consider deprecating `unit_x` etc in favor of these new constants.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog], and this project adheres to
 [Semantic Versioning].
 
 ## [Unreleased]
+### Added
+* Added `Vec2::X`, `Vec4::W` etc as a shorter versions of `unit_x()` and friends.
 
 ### Added
 
--- a/src/vec2.rs
+++ b/src/vec2.rs
@@ -9,6 +9,12 @@ use std::iter::{Product, Sum};
 
 macro_rules! impl_vec2_common_methods {
     ($t:ty, $vec2:ident, $vec3:ident, $mask:ident, $inner:ident) => {
+        /// `[1, 0]`: a unit-length vector pointing along the positive X axis.
+        pub const X: Self = Self($inner::UNIT_X);
+
+        /// `[0, 1]`: a unit-length vector pointing along the positive Y axis.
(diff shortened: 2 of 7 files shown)


bitshifter:
Switching to constants makes sense, I couldn't do that initially since I didn't have a way of making it work with SSE.

I'm not totally sold on `X` vs `UNIT_X`. `X`, `Y` & `Z` are the names of the basis vectors but basis vectors aren't necessarily unit length. On the other hand I am a fan brevity, and people would work it out I guess. Is this a convention you are used to in math libraries? I've seen constants named `I`, `J` and `K` before but that was more confusing.

emilk:
[Standard basis vectors](https://en.wikipedia.org/wiki/Standard_basis) are always unit length, and I don't think any `glam` user would suspect `Vec3::X` to have any other value than `[1, 0, 0]`, so I don't anticipate any confusion there. The only possible source of confusion would be having both `X` *and* `UNIT_X`, so I agree we should go with one and stick to it.

As for conventions: I take my inspiration from math where it is common to represent the standard basis vector with e.g. `x̄`, `ȳ` etc. So `5x̄ + 3ȳ` is equivalent with `[5, 3]ᵀ` (though you also see alternatives like `eₓ`). In this context any length is defined by the basis vectors, so by definition `x̄` is unit length (because it *is* the unit).

As for other math libraries: my personal preference is to design interfaces more from "what is logical and simple" and less from "what do others do" :) (but of course I see the advantage of being consistent).

bitshifter:
Alright, I'll merge these. For consistency it might be good to add `ZERO`, `ONE` and `IDENTITY` and eventually deprecate and remove the method versions.

PR #124, opened 2021-01-26
Add ZERO, ONE and IDENTITY constants

emilk:
Add constants for `ZERO` for vectors and matrices, `ONE` for vectors, and `IDENTITY` for matrices and quaternions.

Deprecates the old `zero()`, `one()` and `identity()` functions.

This is a continuation of https://github.com/bitshifter/glam-rs/pull/120 and especially https://github.com/bitshifter/glam-rs/pull/120#issuecomment-761885989

I omited `ZERO` for quaternion because there were no `zero()` function, and also because I think there is less use for it (most uses of quaternions keep them normalized), which is probably why there was no `zero()` function.

Let me know if you think it is too early to add deprecation notices!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,14 @@ The format is based on [Keep a Changelog], and this project adheres to
 ## [Unreleased]
 ### Added
 * Added `Vec2::X`, `Vec4::W` etc as a shorter versions of `unit_x()` and friends.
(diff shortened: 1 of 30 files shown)


bitshifter:
Hey, I'm away for a few weeks. I'll look at this when I get back and am near a computer.

PR #127, opened 2021-02-05
Add Vec::normalize_or_zero()

emilk:
See discussion in https://github.com/bitshifter/glam-rs/issues/74

This adds the method `normalize_or_zero()` which will either return a unit-length vector or zero, but never a non-finite vector.

On my MacBook Pro 2019 ~I get `4.6 ns` for `Vec3::normalize()` and `8.1 ns` for `Vec3::normalize_or_zero()`.~
**EDIT**: `Vec3::normalize_or_zero()` takes about 40% more time compared to `Vec3::normalize()`.

## Motivation
In many real-world cases, the safe choice is zero when a vector cannot be normalized (regardless of the reason). This PR makes this common use case very ergonomic.

## Alternatives
Instead of normalizing and checking if all values are finite we could take the length, check if it is small or non-finite, and if so return zero. This has the potential of being slightly faster, but the cutoff for "small" must be chosen very carefully.

We could also consider splitting this into two functions: `v.normalize().finite_or_zero()`, i.e. have a special function that returns the zero vector for non-finite vectors. This choice would make it slightly longer than `normalize_or_zero()` and also preclude future optimizations as suggested above. That being said, a `finite_or_zero` might still be a nice separate addition (for another PR).

### DIFF
--- a/src/vec.rs
+++ b/src/vec.rs
@@ -263,7 +263,7 @@ macro_rules! impl_vecn_float_methods {
             if v.is_finite() {
                 v
             } else {
-                Self::default()
+                Self::ZERO
             }
         }
 
--- a/src/vec.rs
+++ b/src/vec.rs
@@ -259,9 +259,9 @@ macro_rules! impl_vecn_float_methods {
         /// the result of this operation will be zero.
         #[inline(always)]
         pub fn normalize_or_zero(self) -> Self {
-            let v = self.normalize();
-            if v.is_finite() {
-                v
+            let rcp = self.length_recip();
+            if rcp.is_finite() {
+                self * rcp
             } else {
                 Self::ZERO
             }
--- a/src/vec.rs
+++ b/src/vec.rs
@@ -260,7 +260,7 @@ macro_rules! impl_vecn_float_methods {
         #[inline(always)]
         pub fn normalize_or_zero(self) -> Self {
             let rcp = self.length_recip();
-            if rcp.is_finite() {
+            if rcp.is_finite() && rcp > 0.0 {
                 self * rcp
             } else {
                 Self::ZERO
--- a/src/vec.rs
+++ b/src/vec.rs
@@ -246,10 +246,25 @@ macro_rules! impl_vecn_float_methods {
 
         /// Returns `self` normalized to length 1.0.
         ///
-        /// For valid results, `self` must _not_ be of length zero.
+        /// For valid results, `self` must _not_ be of length zero, nor very close to zero.
+        /// You can check the results with [`Self::is_finite`], or use [`Self::normalize_or_zero`].
         #[inline(always)]
         pub fn normalize(self) -> Self {
             Self($flttrait::normalize(self.0))
+        }
+
+        /// Returns `self` normalized to length 1.0 if possible, else returns zero.
+        ///
+        /// In particular, if the input is zero (or very close to zero), or non-finite,
+        /// the result of this operation will be zero.
+        #[inline(always)]
+        pub fn normalize_or_zero(self) -> Self {
+            let v = self.normalize();
+            if v.is_finite() {
+                v
+            } else {
+                Self::default()
+            }
         }
 
         /// Returns whether `self` is length `1.0` or not.
--- a/benches/vec3.rs
+++ b/benches/vec3.rs
@@ -97,6 +97,30 @@ op => vec3_into_tuple,
 from => random_vec3
 );
 
+#[inline]
+fn vec3_normalize(v: Vec3) -> Vec3 {
+    v.normalize()
+}
+
+bench_func!(
+    vec3_normalize_bench,
+    "vec3 normalize",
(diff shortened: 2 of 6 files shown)


src/vec.rs
```diff
@@ -244,12 +244,27 @@ macro_rules! impl_vecn_float_methods {
...
+        ///
+        /// In particular, if the input is zero (or very close to zero), or non-finite,
+        /// the result of this operation will be zero.
+        #[inline(always)]
+        pub fn normalize_or_zero(self) -> Self {
+            let v = self.normalize();
+            if v.is_finite() {
+                v
+            } else {
+                Self::default()
```
bitshifter, line 264: Probably doesn't matter hugely, but this could use `::ZERO`.

src/vec.rs
```diff
@@ -244,12 +244,27 @@ macro_rules! impl_vecn_float_methods {
...
         }
 
+        /// Returns `self` normalized to length 1.0 if possible, else returns zero.
+        ///
+        /// In particular, if the input is zero (or very close to zero), or non-finite,
+        /// the result of this operation will be zero.
+        #[inline(always)]
+        pub fn normalize_or_zero(self) -> Self {
+            let v = self.normalize();
+            if v.is_finite() {
```
bitshifter, line 261: You could probably make this cheaper by checking if the reciprocal length is finite rather than every element of the resulting vector. The default trait implementation of normalize is:

```
fn normalize(self) -> Self {
    self.mul_scalar(self.length_recip())
}
```
Only checking rcp would be.
```
fn normalize_or_zero(self) -> Self {
     let rcp = self.length_recip();
     if rcp.is_finite() {
         self.mul_scalar(rcp)
    } else {
        Self::ZERO
    }
}
```
Perhaps that could still miss degenerate results though?

bitshifter:
@emilk I don't know if github notified you but it looks like this PR has a conflict after I merged your other PR.

src/vec.rs
emilk: I expected it to fail for an infinite vector, but surprisingly `Vec2::new(INFINITY, 0.0).length()` is `NAN` rather than `INFINITY`, and `Vec2::new(INFINITY, 0.0).length_recip()` is `NAN` rather than zero. I'm not sure if that will be true on all platforms though (maybe it is a simd thing?).

`if rcp.is_finite() && rcp > 0.0` may be a safer test

emilk:
With the `length_recip` optimization `normalize()` is only 20% slower than `normalize()`

src/vec.rs
```diff
@@ -246,12 +246,27 @@ macro_rules! impl_vecn_float_methods {
...
 
+        /// Returns `self` normalized to length 1.0 if possible, else returns zero.
+        ///
+        /// In particular, if the input is zero (or very close to zero), or non-finite,
+        /// the result of this operation will be zero.
+        #[inline(always)]
+        pub fn normalize_or_zero(self) -> Self {
+            let rcp = self.length_recip();
+            if rcp.is_finite() {
+                self * rcp
```
emilk, line 264: This works because `length_recip()` returns `NAN` for infinite vectors. Can we trust that behavior is cross platform? Does the CI tests both SIMD on and SIMD off?

bitshifter: I would expect it's IEEE behaviour so it should be cross platform. I do test with and without simd enabled. Vec2 and Vec3 won't be using simd in any case.

emilk: Oh wait, I found the problem. I was thinking of `INFINITY * Vec2::X` as `Vec2::new(INFINITY, 0.0)`, but the are not equivalent. I will fix my tests

src/vec.rs
emilk: Nevermind that comment - my tests where wrong. `Vec2::new(INFINITY, 0.0).length()` is INFINITY.

The `if rcp.is_finite()` version does NOT work for infinite vectors, but `if rcp.is_finite() && rcp > 0.0` works for all my tests, and is faster than my original version. On my machine the original `normalize + is_finite` version was 80% slower than vanilla `normalize`, but the new `rcp.is_finite() && rcp > 0.0` is only around 40% slower than vanilla `normalize`

emilk:
Ok, I think this is good to go now finally :)

bitshifter:
I can merge this as is although I am beginning to wonder if we should go with testing squared length against some epsilon  and not use `is_finite`. The reason being dividing by zero before testing will trigger floating point exceptions, which I often use on work projects to try track down NaNs (in C++ projects, but `f32` is not radically different in Rust). Because this method will divides by zero and then check the result it would trigger that exception which would be annoying if debugging with float exceptions.

emilk:
Well, how the function in implemented can always be changed later. I think what’s more important is to nail down the name and the semantics of it. In particular, I think it is a good idea to never return a non-finite value from `normalize_or_zero`, even when the input is infinite or NaN.

PR #128, opened 2021-02-10
Add try_normalize for all real vectors

emilk:
This is a follow-up to https://github.com/bitshifter/glam-rs/pull/127 and https://github.com/bitshifter/glam-rs/issues/74

`fn try_normalize(&self) -> Option<Self>` is a natural choice when you either:

* Want to early-out on invalid vectors: `v.try_normalize()?`
* Want to do proper error handling: `v.try_normalize.ok_or(BadInputError)?`
* Do hard panics on invalid input: `v.try_normalize().unwrap()`

### DIFF
--- a/src/vec.rs
+++ b/src/vec.rs
@@ -254,10 +254,10 @@ macro_rules! impl_vecn_float_methods {
             Self($flttrait::normalize(self.0))
         }
 
-        /// Returns `self` normalized to length 1.0 if possible, else returns None.
+        /// Returns `self` normalized to length 1.0 if possible, else returns `None`.
         ///
         /// In particular, if the input is zero (or very close to zero), or non-finite,
-        /// the result of this operation will be None.
+        /// the result of this operation will be `None`.
         ///
         /// See also [`Self::normalize_or_zero`].
         #[inline]
--- a/src/vec.rs
+++ b/src/vec.rs
@@ -278,7 +278,12 @@ macro_rules! impl_vecn_float_methods {
         /// See also [`Self::try_normalize`].
         #[inline]
         pub fn normalize_or_zero(self) -> Self {
-            self.try_normalize().unwrap_or(Self::ZERO)
+            let rcp = self.length_recip();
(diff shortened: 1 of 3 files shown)


src/vec.rs
```diff
@@ -247,26 +247,40 @@ macro_rules! impl_vecn_float_methods {
...
 
+        /// Returns `self` normalized to length 1.0 if possible, else returns zero.
+        ///
+        /// In particular, if the input is zero (or very close to zero), or non-finite,
+        /// the result of this operation will be zero.
+        ///
+        /// See also [`Self::try_normalize`].
+        #[inline]
+        pub fn normalize_or_zero(self) -> Self {
+            self.try_normalize().unwrap_or(Self::ZERO)
```
bitshifter, line 281: Using `try_normalize` here adds overhead which isn't really necessary, we could live with some code duplication to avoid using `Option` here.

src/vec.rs
```diff
@@ -247,26 +247,40 @@ macro_rules! impl_vecn_float_methods {
...
-        /// You can check the results with [`Self::is_finite`], or use [`Self::normalize_or_zero`].
+        ///
+        /// See also [`Self::try_normalize`] and [`Self::normalize_or_zero`].
         #[inline(always)]
         pub fn normalize(self) -> Self {
             Self($flttrait::normalize(self.0))
         }
 
-        /// Returns `self` normalized to length 1.0 if possible, else returns zero.
+        /// Returns `self` normalized to length 1.0 if possible, else returns None.
```
bitshifter, line 257: None should probably be `None`

src/vec.rs
```diff
@@ -247,26 +247,40 @@ macro_rules! impl_vecn_float_methods {
...
         ///
         /// In particular, if the input is zero (or very close to zero), or non-finite,
-        /// the result of this operation will be zero.
-        #[inline(always)]
-        pub fn normalize_or_zero(self) -> Self {
+        /// the result of this operation will be None.
```
bitshifter, line 260: `None`

Issue #76, opened 2020-09-04
Support for const types


ApoorvaJ:
Currently, constant types don't seem to be supported:

```
error[E0015]: calls in constants are limited to constant functions, tuple structs and tuple variants
  --> src/main.rs:
   |
17 | const INIT_POS: Vec3 = vec3(1.0, 1.0, 1.0);
   |                        ^^^^^^^^^^^^^^^^^^^
```
I can't construct the tuple manually either, since there is only crate-level visibility. Would it be possible to implement this?

bitshifter:
I can't do this across the board because the intrinsics that glam uses are not const fn. I'm a bit reluctant to do it to some types and not others because it would be inconsistent which could be confusing. As soon as intrinsics get const fn I would be keen to use it.

In the interim you could use an array or tuple are truly in the into/from implementations.

bitshifter:
I guess I could do some union hacks to avoid calling the intrinsic functions. Last time I played around with doing that it was a bit of a perf hit so it might not pan out, should be possible though.

ApoorvaJ:
It's definitely not a pressing issue. I can easily work around it. I really like glam for its simplicity and uniformity. Just wanted to check whether this was an oversight or a limitation. :) Feel free to close this issue as you see fit.

bitshifter:
No worries, it's something I would like to address. Will leave this open and try out the union cast thing some time.

bitshifter:
I tried this out but unfortunately performing union casts and other unsafe casts is not allowed in `const fn`, tracking issue is here https://github.com/rust-lang/rust/issues/51909.

It might be a while until Rust gets that feature but I'll leave this open for now anyway.

FSMaxB:
Rather than making operations `const fn`, would it be possible to make just the constructors `const fn`?

bitshifter:
That is what I was attempting to do. The simd types like __m128 are opaque and you need to call functions like _mm_set_ps to initialise them. Those functions are not const. I don't know if there's any reason those std lib intrinsic functions couldn't be made const though.

FSMaxB:
> That is what I was attempting to do.

Oh, ok.

bitshifter:
I raised an issue with `stdarch` about this. Let's see where it goes. https://github.com/rust-lang/stdarch/issues/906

bitshifter:
It looks like it should be possible to make the constructors in `std::arch` `const fn`. That might take a while do though.

[...]

PR #129, opened 2021-02-10
Quat::from_rotation_arc

emilk:
Here are some very useful functions! In particular `Quat::from_rotation_arc` for when you want the quaternion that rotates one vector to another.

### DIFF
--- a/src/quat.rs
+++ b/src/quat.rs
@@ -29,7 +29,7 @@ use core::{
 use std::iter::{Product, Sum};
 
 macro_rules! impl_quat_methods {
-    ($t:ty, $quat:ident, $vec3:ident, $mat3:ident, $mat4:ident, $inner:ident) => {
+    ($t:ident, $quat:ident, $vec3:ident, $mat3:ident, $mat4:ident, $inner:ident) => {
         /// The identity quaternion. Corresponds to no rotation.
         pub const IDENTITY: Self = Self($inner::W);
 
@@ -145,8 +145,8 @@ macro_rules! impl_quat_methods {
                 Self::from_xyzw(v.x, v.y, v.z, w2 / 2.0)
             } else {
                 // 180° singulary.
-                let tau = 6.28318530717958647692528676655900577;
-                Self::from_axis_angle(from.any_orthonormal(), tau / 2.0)
+                let pi = std::$t::consts::PI; // half a turn = 𝛕/2 = 180°
+                Self::from_axis_angle(from.any_orthonormal(), pi)
             }
         }
 
--- a/src/quat.rs
+++ b/src/quat.rs
@@ -133,19 +133,19 @@ macro_rules! impl_quat_methods {
         /// The rotation is in the plane spanned by the two vectors.
         /// Will rotate at most 180 degrees.
         ///
-        /// The input vectors must be finite and non-zero.
+        /// The input vectors must be normalized (unit-length).
         ///
         /// `from_rotation_arc(from, to) * from ≈ to`.
         pub fn from_rotation_arc(from: $vec3, to: $vec3) -> Self {
-            let from = from.normalize();
-            let to = to.normalize();
-            let w2 = (2.0 + 2.0 * from.dot(to)).sqrt();
-            let w2_inv = 1.0 / w2;
-            if w2_inv.is_finite() {
-                let v = from.cross(to) * w2_inv;
+            glam_assert!(from.is_normalized());
+            glam_assert!(to.is_normalized());
+            let w2_sqr = (2.0 + 2.0 * from.dot(to));
+            if w2_sqr > 16.0 * $t::EPSILON {
+                let w2 = w2_sqr.sqrt();
+                let v = from.cross(to) / w2;
                 Self::from_xyzw(v.x, v.y, v.z, w2 / 2.0)
             } else {
-                // 180° singulary.
+                // 180° singulary: from ≈ -to
                 let pi = std::$t::consts::PI; // half a turn = 𝛕/2 = 180°
                 Self::from_axis_angle(from.any_orthonormal(), pi)
             }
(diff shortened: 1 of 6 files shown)


src/quat.rs
```diff
@@ -128,6 +128,45 @@ macro_rules! impl_quat_methods {
...
+            let from = from.normalize();
+            let to = to.normalize();
+            let w2 = (2.0 + 2.0 * from.dot(to)).sqrt();
+            let w2_inv = 1.0 / w2;
+            if w2_inv.is_finite() {
+                let v = from.cross(to) * w2_inv;
+                Self::from_xyzw(v.x, v.y, v.z, w2 / 2.0)
+            } else {
+                // 180° singulary.
+                let tau = 6.28318530717958647692528676655900577;
```
emilk, line 148: didn't quite know how to get in `std::f32::consts::TAU` here in a nicer way, so I just hard-coded it 🤷

tests/vec3.rs
```diff
@@ -457,6 +457,8 @@ macro_rules! impl_vec3_float_tests {
         use core::$t::NAN;
         use core::$t::NEG_INFINITY;
 
+        const NONZERO_TEST_VECTORS: [$vec3; 3] = [$vec3::X, $vec3::Y, $vec3::Z];
```
emilk, line 460: I did plan to add some better vectors, but was foiled by `$vec3::new` [not being const](https://github.com/bitshifter/glam-rs/issues/76). I could make a function return these instead, or just inline them, but I put it like this as a request for suggestions :)

src/quat.rs
bitshifter: `$t::consts::TAU`? I think it's pretty recent though, might not be worth bumping MSRV for.

tests/vec3.rs
bitshifter: There are macros to work around that, e.g. `const_vec3!`. Probably the easiest thing to do is to set up the constants in the `vec3` or `dvec3` scope and then reference them from the common macros that add the tests.

src/quat.rs
bitshifter: If you are using `TAU / 2.0` why not use `PI`?

src/quat.rs
```diff
@@ -128,6 +128,45 @@ macro_rules! impl_quat_methods {
...
+        /// Gets the minimal rotation for transforming `from` to `to`.
+        /// The rotation is in the plane spanned by the two vectors.
+        /// Will rotate at most 180 degrees.
+        ///
+        /// The input vectors must be finite and non-zero.
+        ///
+        /// `from_rotation_arc(from, to) * from ≈ to`.
+        pub fn from_rotation_arc(from: $vec3, to: $vec3) -> Self {
+            let from = from.normalize();
+            let to = to.normalize();
```
bitshifter, line 140: Typically in `glam` I have avoided normalizing input vectors and instead expect the input to be pre-normalized. I have `glam_assert` which can be enabled to check that parameters are valid if necessary. I try and comment where inputs need to meet some requirement like this although sometimes I've forgetten to add a comment (e.g. Quat::from_axis_angle`. I work on the premise that people shouldn't pay the cost if their data is already normalized. Often calling code may also need to normalize their vectors for whatever else they are doing, so I figure if I require parameters to be normalized then it won't happen twice.

I suppose in this case it would be reasonable for vectors to not necessarily be unit length...

Alternatively, this could optionally normalize if the input is not already normalized, which at least would avoid normalizing already normalized data.

src/vec3.rs
```diff
@@ -193,6 +193,28 @@ macro_rules! impl_vec3_float_methods {
...
         }
+
+        /// Returns somes vector that is orthogonal to the given one.
+        ///
+        /// The output vector is not necessarily unit-length.
+        /// For that use [`Self::any_orthonormal`] instead.
+        ///
+        /// The input vector must be finite and non-zero.
+        #[inline]
+        pub fn any_orthogonal(&self) -> Self {
```
bitshifter, line 204: I haven't come across these before, I'm curious what they are used for?

src/quat.rs
emilk: Heh, habits :) I always think of angles in terms of 𝛕

src/vec3.rs
emilk: For the `from_rotation_arc` I use it to pick an axis of rotation in the case of a 180° singularity. That is, to rotate from `+v` to `-v` you can pick any axis that is orthogonal to `v`. 

There are many other cases where it is useful to be able to create an orthonormal basis. For instance you may have a surface normal but want to be able to create an ad-hoc tangent space, so you need a tangent. There are an infnite number of tangents to pick from, so: `let tangent = normal.any_orthonormal();`.

My implementation here isn't the most efficient though. I should probably try out this method instead: https://graphics.pixar.com/library/OrthonormalB/paper.pdf (especially for the `any_orthonormal()` method).

If it's worth doing, it's worth doing right. I'll make some additional commits improving this (and add a benchmark)

src/quat.rs
emilk: I struggled with `$t::consts::TAU` but now realized I needed to change `$t:ty` to `$t:ident`.

I changed this to `std::$t::consts::PI` now

tests/vec3.rs
emilk: But then I need to duplicate the test vectors (for both Vec3 and DVec3)? I think I'll just define the vector in the test function for now

src/quat.rs
emilk: I think your design choice "pay for what you use" makes sense. In this case I think it is reasonable to expect users to normalize themselves. I'll replace the normalization with `glam_assert` and a docstring!

src/vec3.rs
bitshifter: Cool, thanks for the explanation 👍

src/quat.rs
```diff
@@ -78,6 +78,7 @@ macro_rules! impl_quat_methods {
         }
 
         /// Create a quaterion for a normalized rotation axis and angle (in radians).
+        /// The axis must be normalized (unit-length).
```
bitshifter, line 81: thanks :)

tests/vec3.rs
emilk: I added a `vec3_float_test_vectors!()` macro

emilk:
This became a little bit more work and a bigger PR than I had anticipated, but I think it is done now :)

bitshifter (approved):

PR #130, opened 2021-02-12
Add Quat::angle_between

emilk:
This is very similar to the existing `Vec3::angle_between`


emilk:
I'll cool down with the unsolicited PR:s after this one, I promise!

bitshifter:
Are these all things that you are using at Embark, or from other projects?

bitshifter:
Would be good to add to `CHANGELOG.md` as well

emilk:
These are things we're using at Embark.

bitshifter (approved):

PR #134, opened 2021-02-28
Impl Eq and Hash for ivec and uvec

alec-deason:
This implements Eq and Hash for integer Vec2/3/4 which is a very convenient pair of traits to have.


bitshifter:
Awesome, thanks for that!

PR #140, opened 2021-03-05
Mat4::orthographic_lh is left-handed not right-handed

CAD97:
This definitely confused me for a beat when first reading through the docs 😅


CAD97:
Also went ahead and fixed a couple other typos in the same documentation block.

bitshifter:
Ah whoops, thanks!

PR #142, opened 2021-03-08
Add spirv module

XAMPPRocky:
This PR adds a spirv module responsible for the integration between `glam` and SPIR-V targets, and also adds dependencies on `spirv-std` and moves the trait implementations previously contained in `spirv-std` into `glam`. This is part of a series changes to allow `spirv-std` to be compatible with multiple vector libraries, and to allow libraries like glam to be able to take advantage of intrinsics provided by `spirv-std` such as using vector operations for add, multiply, etc.

This PR is currently blocked on https://github.com/EmbarkStudios/rust-gpu/pull/476 landing so that we can publish a version of `spirv-std` with those changes and prevent cyclic dependencies issues.

### DIFF
--- a/src/vec3.rs
+++ b/src/vec3.rs
@@ -6,6 +6,9 @@ use crate::{BVec3, DVec2, DVec4, IVec2, IVec4, UVec2, UVec4, Vec2, Vec4, XYZ};
 use core::fmt;
 use core::{cmp::Ordering, f32, ops::*};
 
+#[cfg(not(feature = "std"))]
+use num_traits::Float;
+
 #[cfg(all(
     target_arch = "x86",
     target_feature = "sse2",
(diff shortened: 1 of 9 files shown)


src/vec3.rs
```diff
@@ -224,6 +227,7 @@ macro_rules! impl_vec3_float_methods {
         /// Given a unit-length vector return two other vectors that together form an orthonormal basis.
         /// That is, all three vectors are orthogonal to each other and are normalized.
         #[inline]
+        #[cfg(not(target_arch="spirv"))]
```
bitshifter, line 230: How come these ones need to be disabled? Missing copysign?

XAMPPRocky: Yep, copysign isn't available in `num-traits` at the moment. I'm pushing for it to get merged but until then I needed to disable these functions.

PR #146, opened 2021-03-10
Add vector clamp function

emilk:
This mirrors the new https://doc.rust-lang.org/std/primitive.f32.html#method.clamp function, with the differences:

* `assert!` has been replaced with `glam_assert!`
* treatment of `NaN` is undefined

### DIFF
--- a/src/core/scalar/vector.rs
+++ b/src/core/scalar/vector.rs
@@ -232,6 +232,16 @@ impl<T: NumEx> Vector<T> for XY<T> {
             y: self.y.max(other.y),
         }
     }
+
+    #[inline]
+    fn clamp(self, min: Self, max: Self) -> Self {
+        glam_assert!(min.x <= max.x);
+        glam_assert!(min.y <= max.y);
+        Self {
+            x: self.x.max(min.x).min(max.x), // TODO: use std::f32::clamp when we update the minimal required rustc version to >= 1.50
+            y: self.y.max(min.y).min(max.y), // TODO: use std::f32::clamp when we update the minimal required rustc version to >= 1.50
+        }
+    }
 }
 
 impl<T: NumEx> Vector<T> for XYZ<T> {
@@ -383,6 +393,18 @@ impl<T: NumEx> Vector<T> for XYZ<T> {
             x: self.x.max(other.x),
             y: self.y.max(other.y),
             z: self.z.max(other.z),
+        }
+    }
+
+    #[inline]
+    fn clamp(self, min: Self, max: Self) -> Self {
+        glam_assert!(min.x <= max.x);
+        glam_assert!(min.y <= max.y);
+        glam_assert!(min.z <= max.z);
+        Self {
+            x: self.x.max(min.x).min(max.x), // TODO: use std::f32::clamp when we update the minimal required rustc version to >= 1.50
+            y: self.y.max(min.y).min(max.y), // TODO: use std::f32::clamp when we update the minimal required rustc version to >= 1.50
+            z: self.z.max(min.z).min(max.z), // TODO: use std::f32::clamp when we update the minimal required rustc version to >= 1.50
         }
     }
 }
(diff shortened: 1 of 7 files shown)


src/core/scalar/vector.rs
```diff
@@ -232,6 +232,16 @@ impl<T: NumEx> Vector<T> for XY<T> {
             y: self.y.max(other.y),
         }
     }
+
+    #[inline]
+    fn clamp(self, min: Self, max: Self) -> Self {
+        glam_assert!(min.x <= max.x);
+        glam_assert!(min.y <= max.y);
+        Self {
+            x: self.x.max(min.x).min(max.x), // TODO: use std::f32::clamp when we update the minimal required rustc version to >= 1.50
```
bitshifter, line 241: Possibly we don't want to use Rust clamp long term as it will definitely assert/panic. If not following the convention of Rust's clamp it would be worth calling out in the method docs.

src/core/sse2/vector.rs
```diff
@@ -261,6 +261,11 @@ impl Vector<f32> for __m128 {
     fn max(self, other: Self) -> Self {
         unsafe { _mm_max_ps(self, other) }
     }
+
+    #[inline(always)]
+    fn clamp(self, min: Self, max: Self) -> Self {
+        self.max(min).min(max)
```
bitshifter, line 267: This version should also have a `glam_assert` for consistency, e.g. `glam_assert!(min.cmp_le(max).all())`

bitshifter (changes requested):
Mostly good, just a question of if we want to follow Rust's `clamp` semantics or not. Some methods I have followed the standard library (`signum` I think) but I don't think it's totally necessary as long as we call it out.

emilk:
Should we define the behavior of a `NaN` in self,min,max ? Should we protect against that with `glam_assert!`?

src/core/sse2/vector.rs
bitshifter: Hmm, the assert actually complicates things a bit it turns out. For the clamp there's no Vector3/Vector4 implementation required because we just ignore lanes we don't care about, but for the assert we do care about the lanes, or rather we don't want say the w lane to factor into the Vec3 clamp assert. You could remove it for now and I'll deal with it later. Adding `Vector3` and `Vector4` implementations, instead of `Vector` might fix it.

emilk:
`glam_assert!(min.cmple(max).all());` is failing to compile because I put `clamp` in the `Vector` trait, and `.all()` is in `MaskVector3` and `MaskVector4` and rustc doesn't know which one to use. How would you suggest fixing that?

bitshifter:
I made a comment on this but probably in the wrong place.

The assert complicates things a bit it turns out. For the clamp there's no Vector3/Vector4 implementation required because we just ignore lanes we don't care about (e.g. w for vec3), but for the assert we do care about the lanes, or rather we don't want say the w lane to factor into the Vec3 min/max assert. You could remove it for now and I'll deal with it later. I think to fix it. adding Vector3 and Vector4 implementations of clamp instead of Vector might fix it. It might still require the trait for all to be specified explicitly. Like I said, I'm happy for you to remove the assert and let me deal with it also.

src/core/sse2/vector.rs
```diff
@@ -261,6 +261,15 @@ impl Vector<f32> for __m128 {
     fn max(self, other: Self) -> Self {
         unsafe { _mm_max_ps(self, other) }
     }
+
+    #[inline(always)]
+    fn clamp(self, min: Self, max: Self) -> Self {
+        glam_assert!(
+            MaskVector3::all(min.cmple(max)), // TODO: also check .w for Vec4
+            "clamp: expected min <= max"
+        );
```
emilk, line 270: I think this is a reasonable compromise for now, checking XYZ for both Vec3 and Vec4, but missing W for Vec4.

PR #151, opened 2021-03-25
Update spirv-std

XAMPPRocky:


XAMPPRocky:
Seems like this is hitting unrelated clippy failures in the newest Rust version.

bitshifter:
Wow, that new lint is annoying... I'll look into it.

PR #152, opened 2021-03-26
Update spirv-std to 0.4.0-alpha.3

XAMPPRocky:
Additional changes were needed to be published. 🙂

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,9 +42,7 @@ mint = { version = "0.5", optional = true, default-features = false }
 num-traits = { version = "0.2.14", optional = true, default-features = false }
 rand = { version = "0.7", optional = true, default-features = false }
 serde = { version = "1.0", optional = true, features = ["derive"] }
-
-[target.'cfg(target_arch = "spirv")'.dependencies]
-spirv-std = { version = "0.4.0-alpha.1", features = ["const-generics"] }
+spirv-std = { version = "0.4.0-alpha.3", optional = true }
 
 [dev-dependencies]
 criterion = "0.3"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu, macos, windows]
-        rust: [1.36.0, stable, beta, nightly]
+        rust: [1.45.0, stable, beta, nightly]
     runs-on: ${{ matrix.os }}-latest
     steps:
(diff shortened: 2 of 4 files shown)


bitshifter:
I've fixed the lint problem on master, but it looks like there is a different issues with this PR still.

bitshifter:
Seems like `rustup default stable` is still getting 1.50.0. I've had problems with the stable Rust version being cached before on travis but not github. Not sure how to force it to updates...

There are some other issues with the 1.36.0 build, Is that something that can easily be addressed in the SPIR-V crate? I'm happy to not check MSVR for SPIR-V stuff, although this is in a regular x86 build so not sure what's going on there.

XAMPPRocky:
> There are some other issues with the 1.36.0 build, Is that something that can easily be addressed in the SPIR-V crate? I'm happy to not check MSVR for SPIR-V stuff, although this is in a regular x86 build so not sure what's going on there.

Yeah, it should be possible to patch around that so it works on 1.36.0. I'll. work on that, though it'll require another version 😄  It's happening because `spirv-std-macros` is a proc macro crate, and it's using features not available for compilers that old.

bitshifter:
What Rust version does it require? I'll have to bump MSVR sometime.

bitshifter:
If you want to bump the msvr to 1.42 or 1.45 as part of the pr that's fine. I'll make a major release  for a msvr change but I'd be surprised if anyone complains.

bitshifter:
This is just github actions not using latest Rust again...

bitshifter:
Try rebasing with latest, I think I've forced the CI to update stable.

Cargo.toml
```diff
@@ -42,9 +42,7 @@ mint = { version = "0.5", optional = true, default-features = false }
 num-traits = { version = "0.2.14", optional = true, default-features = false }
 rand = { version = "0.7", optional = true, default-features = false }
 serde = { version = "1.0", optional = true, features = ["derive"] }
-
-[target.'cfg(target_arch = "spirv")'.dependencies]
-spirv-std = { version = "0.4.0-alpha.1", features = ["const-generics"] }
+spirv-std = { version = "0.4.0-alpha.3" }
```
bitshifter, line 45: I don't really want to make this a default dependency for all glam users as it's only relevant for rust-gpu. I assume you want it on non SPIR-V targets though? Perhaps it could be enabled via a feature if that was the case?

bitshifter (changes requested):

bitshifter (approved):

PR #154, opened 2021-03-31
Add Vec2::AXES, Vec3::AXES, Vec4::AXES

emilk:
This is useful when writing code that is generic over the dimensions:

``` rust
for dim in 0..3 {
    let axis = Vec3::from_axis(dim);
    …
}
```

I'm open to suggestions for the name though. `from_unit_dimension`, `from_numbered_axis` etc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,18 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog], and this project adheres to
 [Semantic Versioning].
 
+## Unreleased
+
+### Added
+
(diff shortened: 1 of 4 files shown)


bitshifter:
Perhaps these would be better as constants? For example:

```
impl Vec3 {
        pub const AXES: [Self; 3] = [Self::X, Self::Y, Self::Z];
}
```

```
for axis in &Vec3:AXES {
    ....
}
```

What do you think? I feel like this is simpler, not sure what the best name would be though.

emilk:
I like it! Much better! 😃  I think `AXES` is a good name too, and very self-explanatory. I'll fix this up tomorrow.

PR #156, opened 2021-04-01
WIP: Mat3x4

emilk:
Part of https://github.com/bitshifter/glam-rs/issues/25

This was more work than I thought (it always is). I can't say I fully grok how all the different traits fit together, but I tried to copy the basic structure of the 4x4 types. I still have some hard-to-decipher errors.

## TODO:

* [ ] Get it to compile
* [ ] Finish up the TODO:s an commented code
* [ ] Write tests
* [ ] Write benchmarks

Issue #25, opened 2019-09-18
Consider adding transform types


bitshifter:
When dealing with transforms that only contain position, orientation and possibly scale many operations can be performed more efficiently than using a general purpose matrix.

A while ago I experimented with transform types which contained position and orientation (as a quaternion) much like Unreal's `FTransform`, but I found these to not be particularly more efficient or to save much space in practice, so they are not enabled by default. They can be enabled using the `transform-types` feature.

Another option might be to have these kind of transforms backed by a matrix with simplified methods/operators similar to `Transform4D` described in Foundations of Game Engine Development Volume 1.

cessen:
Somehow I missed this issue when I was checking for existing issues before filing #62.  I think #62 can be merged with this issue, as it falls under the umbrella of the title.

cessen:
Quoted from issue #62:

> I've been poking at Psychopath a bit again, and I realized that in the context of a ray tracer you pretty much never need full projective transforms, you only need affine transforms.  And this presents a lot of opportunities for optimization:
>
> * An affine transform matrix only needs to store 4x3 floats, not 4x4, because the fourth row is implicitly known.
> * Transforming points/vectors can be further optimized due to the statically-known implicit fourth row.
> * Matrix inversion can also be further optimized.
> * There are tricks for directly performing inverse transforms on points/vectors that are more efficient than fully inverting the matrix + transforming.

To expand on that: affine transforms cover all of the transforms that are useful for object-space/world-space transforms:

* Translation
* Rotation
* Scale
* Shear

The only thing full 4x4 matrices cover beyond that are projective transforms (e.g. perspective projection), which in the context of graphics are typically only used for camera projections to screen space.  So I think affine transforms are useful to have as a separate type: client code can use them whenever object transforms are the concern, and screen space isn't relevant.  It both makes the intent clearer and opens a lot of doors for further optimization as noted above.

Lastly, the current implementation of `Mat4::transform_point3()`, as noted in its comment, is optimized in a way that won't work correctly for perspective projection matrices.  That makes a lot of sense given that the common case is likely to be object transforms, but it's a bit unfortunate that it's not correct for the general case.  Making that compromise is only necessary because `Mat4` is trying to simultaneously serve two different purposes: both affine and projective transforms.  If we add a separate affine transform type, then we won't need that compromise: the affine transform type can have the optimized version, and `Mat4` can be correct for the general case.

cessen:
[...]

PR #157, opened 2021-04-06
Affine3D

emilk:
Part of https://github.com/bitshifter/glam-rs/issues/25. A simpler version of <https://github.com/bitshifter/glam-rs/pull/156>.

This PR introduces the `Affine3D` type, implemented as `3x Vec4` (NOTE: only `f32` version included).

Banchmark results on my Intel MacBook (best of a few runs):

| op  | `Mat4 sse2` | `Affine3D sse2` | | `Mat4 scalar` | `Affine3D scalar` |
| -------------------- | --------- | -------- | - | ------ | -------- |
| `inverse`            |  `12 ns`  | `9 ns` 🥇   | | `31 ns`  | `16 ns` 🥇   |
| `Self * Self`          |  `6.1 ns`  | `4.2 ns` 🥇 | | `21 ns`  | `14 ns` 🥇   |
| `transform point3`   |  `2.6 ns` 🥇  | `3.8 ns` | | `3.8 ns` | `3.5 ns` 🥇  |
| `transform vector3`  |  `2.5 ns` 🥇  | `3.6 ns` | | `3.3 ns`  | `2.9 ns` 🥇 |
 
( 🥇  is the fastest in each pair of columns)

It would be nice to speed up the sse2 vector transforms more, as those are probably among the most common operations to do with a matrix. I've tried several different approaches, but I can't get further than this. Unless someone has a better idea, I think we should just recommend turning your `Affine3D` into a `Mat4` on SSE2 targets when doing a lot of `mat * vec` transforms.

It is also possible that for some platforms (in particular spirv) a `4x Vec3` may actually be faster.


src/mat3x4.rs
```diff
@@ -0,0 +1,984 @@
...
+    /// This is the equivalent of multiplying the 3D vector as a 4D vector where `w` is
+    /// `0.0`.
+    #[inline]
+    #[cfg(not(any(
+        target_arch = "spirv",
+        all(target_feature = "sse2", not(feature = "scalar-math"))
+    )))]
+    pub fn transform_vector3(&self, vec: Vec3) -> Vec3 {
+        Vec3::new(
+            Vec3::from(self.x_row).dot(vec),
```
bitshifter, line 586: If you use `Vec3A` here it will probably perform better.

src/mat3x4.rs
```diff
@@ -0,0 +1,984 @@
...
+#[cfg(all(
+    target_feature = "sse2",
+    not(feature = "scalar-math"),
+    target_arch = "x86_64"
+))]
+use core::arch::x86_64::*;
+
+#[cfg(all(target_feature = "sse2", not(feature = "scalar-math")))]
+#[inline(always)]
+unsafe fn m128_from_vec4(vec4: Vec4) -> __m128 {
```
bitshifter, line 21: There's already a `From` impl to do this, you can use `into()`.

bitshifter:

There are a couple of things I'm not sure about as it currently stands, on naming and being row major.

Naming wise I was thinking of making this a special type, e.g. `Transform3D` or `Affine3D` or something along those lines. There are a few reasons:

* It doesn't tie you to an internal representation, e.g. 4x3 vs 3x4 or rows vs columns
* Methods like `inverse`, `determinant` etc. don't make sense for non square matrices, to be fair I have helper methods on `Mat4` like `transform_point3` etc. that don't make mathematical sense either
* It doesn't make sense to multiply two 3x4 matrices but these transforms be able to be multiplied.
* Making it a special type makes it really clear that it contains a valid affine transform and nothing else

I'm not totally sure about using row major when everything else is column major, it might cause confusion when debugging if the internals behave differently to the other matrix types. For scalar code, the size would be the same for 4x3 or 3x4 and I suspect it wouldn't make a huge difference one way or another for scalar performance. At the moment the SSE2 code isn't benefiting from row major, although I think switching to `Vec3A` in the implementation would get it closer to the `Mat4` performance. Another concern I have is I think accessing the translation and the basis vectors is a very common thing to do and this might be slower if the data is organised in rows, also direct access is nice. Also converting to a `Mat4` or `Mat3` would require a transpose with row major.

src/mat3x4.rs
emilk: No - this is the scalar path, and Vec3A is (slightly) slower (3.1 ns vs 2.9 ns)

src/mat3x4.rs
emilk: Oh :)

emilk:
> Naming wise I was thinking of making this a special type, e.g. `Transform3D` or `Affine3D` or something along those lines.

I agree. I think `Affine3D` is the better name here (since it is more specific).

> I'm not totally sure about using row major when everything else is column major.

I made this choice because I wanted to represent the transform using `12xf32`, to save memory (for better cache locality). If we don't care about that then `4xVec3A` columns makes much more sense. What is "best" choice will depend a lot on the circumstance. The extra saved memory can make a big difference to decrease cache misses, but for applications where you can read data in order, it won't matter much at all.

I can make a competing `4xVec3A` column major implementation so we can compare raw numbers.

emilk:
I think if we rename this `Affine3D` we should also hide all functions related to columns and rows to make it clear that this is a higher abstraction.

I've started work on a 4xVec3A version, but am having some problems. Maybe it can wait a bit and come as a follow-up PR where we switch to a (potentially) faster implementation?

bitshifter:
I think it makes sense to remove those methods. My approach has usually been to provide a pretty minimal amount of functionality and add things that people are requesting.

No problem skipping the 4xVec3A implementation for now if you are having problems with it.

src/mat3x4.rs
bitshifter: Ah my bad, I missed the outer `not`

bitshifter:
One thing I was planning on experimenting with was loading scalar types into SIMD to perform math heavy operations, so it wouldn't just be things like `Vec3A` that are `SIMD` optimised. If doing this improve performance of `Mat3` inverse and vector multiplication functions potentially an `Affine3D` type could just be a composition `Mat3` and `Vec3`. It would be similar in essence to the existing transform type https://github.com/bitshifter/glam-rs/blob/master/src/transform.rs#L217-L224. This does depend a bit on the results of my experiments though :)

The SSE2 dot product is not super fast. The column major vector multiplication doesn't require dot product so it should be a lot faster if  SIMD is available, see https://github.com/bitshifter/glam-rs/blob/master/src/core/sse2/matrix.rs#L406-L421.

bitshifter:
Related to my comment above, I wrote a bit more detail about where I'm thinking of going with glam here https://github.com/bitshifter/glam-rs/discussions/159. You don't really have to do anything with this information for this PR, just thought it might be useful to know what I'm thinking on the performance side of things.

emilk:
Current public interface:

Screen Shot 2021-04-07 at 15 25 33

emilk:
I have a working 4xVec3A implementation now too, but so far it is slower at simd self*self than the 3xVec4 version. It is faster at transforming vectors though. I'll keep tweaking at it, but it can be a separate PR if this gets merged.

src/affine3d.rs
```diff
@@ -0,0 +1,639 @@
...
+    /// The thirs column.
+    #[inline(always)]
+    fn z_col(&self) -> Vec3 {
+        Vec3::new(self.x_row.z, self.y_row.z, self.z_row.z)
+    }
+
+    /// The translation expressed by this transform.
+    /// The translation is applied last, so is separatable from scale, shear and rotation.
+    #[inline(always)]
+    pub fn translation(&self) -> Vec3 {
```
bitshifter, line 285: It might also be good to provide accessors to basis vectors and possibly the rotation/scale 3x3 matrix, although this could happen separate to this PR.

src/features/impl_serde.rs
```diff
@@ -607,3 +607,110 @@ mod u32 {
...
+        ser::{Serialize, SerializeTupleStruct, Serializer},
+    };
+
+    impl Serialize for Affine3D {
+        fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+        where
+            S: Serializer,
+        {
+            // Serialize column-wise as 3x4 matrix:
+            let mat4: Mat4 = (*self).into();
```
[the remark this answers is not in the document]
emilk: Because there are no other ways to access the matrix-form of an `Affine3D`. I could maybe add to/from `(Mat3, Vec3)` pairs ?

src/quat.rs
```diff
@@ -136,6 +136,12 @@ macro_rules! impl_quat_methods {
             ))
         }
 
+        /// From the columns of a 3x3 rotation matrix.
+        #[inline]
+        pub fn from_rotation_axes(x_axis: $vec3, y_axis: $vec3, z_axis: $vec3) -> Self {
```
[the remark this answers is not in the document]
emilk: So I can use it from `Affine3D::to_scale_rotation_translation`

src/affine3d.rs
```diff
@@ -0,0 +1,639 @@
...
+    /// Returns `true` if any elements are `NaN`.
+    #[inline]
+    pub fn is_nan(&self) -> bool {
+        self.x_row.is_nan() || self.y_row.is_nan() || self.z_row.is_nan()
+    }
+
+    /// Returns `true` if all elements are finite and
+    /// this is NOT a degenerate transform (i.e. a transform with a zero scale).
+    #[inline]
+    pub fn is_invertible(&self) -> bool {
```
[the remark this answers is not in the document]
emilk: I agree that a `try_inverse` would be nice for this and all other types, but that's a big task.

I added `is_invertible()` in lieu of `determinant()` which can be used to check for degenerate matrices for e.g. `Mat4`. It can be nice to be able to check for degenerancy (or rather, its opposite) with a convenient method. I could probably have a better name, but I couldn't find it (`is_valid`, `is_nice`, `is_sounds`, ...)

src/affine3d.rs
emilk: I wanted to keep the interface a bit more implementation agnostic for now, but setting/getting the full `Mat3` would be useful.

src/affine3d.rs
```diff
@@ -0,0 +1,639 @@
...
+        }
+    }
+}
+
+impl core::ops::Mul<Mat4> for Affine3D {
+    type Output = Mat4;
+
+    #[inline(always)]
+    fn mul(self, rhs: Mat4) -> Self::Output {
+        Into::<Mat4>::into(self) * rhs
```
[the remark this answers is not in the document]
emilk: You are right; I don't know what I was thinking :)

src/affine3d.rs
emilk: Added setter/getter for `Mat3`

src/affine3d.rs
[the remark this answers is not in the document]
bitshifter: I'd still rather not commit to this in the API at this stage, I'd rather remove it and come up with a more general solution later. If it's absolutely necessary to keep it, then I guess it can be deprecated when an alternative solution is found.

src/quat.rs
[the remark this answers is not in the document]
bitshifter: I'd prefer it was `pub(crate)` for now.

bitshifter:
The weird clippy error was just a run of the mill warning due to the method not being used by the `DQuat` implementation. If you still wanted to use it internally you could make it `#[allow(dead_code)]`.

I've been asked to do a release by another embarker, so I might hold off merging this until I've done that. If you wanted you could put it on a feature so it won't affect the release, or just wait. I want to hold off having this on by default until the column major PR lands. I don't expect it to take long to do a release.

bitshifter:
The release has been done, so should be good to merge once the conflicts are resolved.

emilk:
Rebased to fix merge conflicts in `CHANGELOG.md`

PR #162, opened 2021-04-12
Optimize Affine3D using 4x Vec3A

emilk:
This makes the SSE vector transforms faster, which argueably is the most important. But a lot of other numbers gets slightly worse (including memory usage).

| op                    | old sse2 | new sse2 |   | old scalar | new scalar |
| --------------------- | -------- | -------- | - | ---------- | ---------- |
| `inverse`             | `9.3 ns` | `8.4 ns` 🥇 |   |  `16 ns`   | `16 ns`   |
| `Affine3D * Affine3D` | `4.2 ns` 🥇 | `5.2 ns` |   |  `13 ns`   | `11 ns` 🥇   |
| `transform point3`    | `3.8 ns` | `2.7 ns` 🥇 |   |  `3.5 ns` 🥇  | `4.0 ns`   |
| `transform vector3`   | `3.6 ns` | `2.5 ns` 🥇 |   |  `2.9 ns` 🥇  | `3.6 ns`   |
| size                  | 48 bytes 🥇 | 64 bytes |   | 48 bytes 🥇 | 64 bytes |

### DIFF
--- a/src/affine3d.rs
+++ b/src/affine3d.rs
@@ -1,28 +1,16 @@
-use crate::{Mat3, Mat4, Quat, Vec3, Vec4};
+use crate::{Mat3, Mat4, Quat, Vec3, Vec3A};
 
 #[cfg(not(feature = "std"))]
 use num_traits::Float;
 
-#[cfg(all(
-    target_feature = "sse2",
-    not(feature = "scalar-math"),
-    target_arch = "x86"
-))]
-use core::arch::x86::*;
-#[cfg(all(
-    target_feature = "sse2",
-    not(feature = "scalar-math"),
-    target_arch = "x86_64"
-))]
-use core::arch::x86_64::*;
-
 /// An affine transform, which can represent translation, rotation, scaling and shear.
 #[derive(Copy, Clone, PartialEq)]
 pub struct Affine3D {
-    // TODO: explore different representations, such as 4x Vec3A or Mat3 + Vec3.
-    x_row: Vec4,
-    y_row: Vec4,
-    z_row: Vec4,
+    x_col: Vec3A,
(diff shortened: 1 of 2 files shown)


src/affine3d.rs
```diff
@@ -591,59 +489,56 @@ impl core::ops::Mul for Affine3D {
...
-        all(target_feature = "sse2", not(feature = "scalar-math"))
-    )))]
+    #[cfg(not(all(target_feature = "sse2", not(feature = "scalar-math"))))]
     #[inline(always)]
     fn mul(self, rhs: Affine3D) -> Self::Output {
         Self {
-            x_row: rhs.mul_row_vec4_from_left(self.x_row),
-            y_row: rhs.mul_row_vec4_from_left(self.y_row),
-            z_row: rhs.mul_row_vec4_from_left(self.z_row),
+            x_col: self.transform_vector3(rhs.x_col.into()).into(),
```
bitshifter, line 538: These are going from `Vec3A` to `Vec3` then effectively splatted x, y, and z. Could probably be made faster if you added `transform_vector3a` which is the actual implementation and keeps everything in SIMD, avoids storing the SIMD columns into 3xf32. Same goes for the call to `transform_point3`.

You might be able to get parity with `Affine3D * Affine3D` doing that.

bitshifter:
I guess the scalar multiply is losing out due to the alignment/padding :(

src/affine3d.rs
emilk: This is the scalar path, so sticking to Vec3A is actually slower (I tested it).

bitshifter: Ah my bad, didn't notice it was the scalar path.

Issue #79, opened 2020-09-11
Coordinate system agnostic


emilk:
`glam` is mostly agnostic to coordinate systems and handedness - and I think that's great. When a choice must be made, `glam` often leaves it to the user to pick, like with `orthographic_lh` + `orthographic_rh`.

There is, however, one exception to this rule: `from_rotation_ypr` (found in `Quat`, `Mat3` and `Mat4`). 

On [`Mat3::from_rotation_ypr`](https://docs.rs/glam/0.9.4/glam/f32/struct.Mat3.html#method.from_rotation_ypr) the documentation simply says `Creates a 3x3 rotation matrix from the given Euler angles (in radians).` which is deceptive, as it implies only one way to construct such a matrix from yaw/pitch/roll. 

[`Quat::from_rotation_ypr`](https://docs.rs/glam/0.9.4/glam/f32/struct.Quat.html#method.from_rotation_ypr) is better, and states: "Create a quaternion from the given yaw (around y), pitch (around x) and roll (around z) in radians.". It still neglects to mention the order. Is this `Quat::from_axis_angle(Y, yaw) * Quat::from_axis_angle(X, pitch) * Quat::from_axis_angle(Z, roll)` or `Quat::from_axis_angle(Z, roll) * Quat::from_axis_angle(X, pitch) * Quat::from_axis_angle(Y, yaw)` ?

But the deeper issue is that this implies that we all agree that `+Y` is up. A lot of people use `+Z` as up and may reach for `from_rotation_ypr` and not realize it won't do what they think it will.

## Solutions

A few potential solutions spring to mind:

* **Remove**: Just delete the function and let people explicitly chain `several from_axis_angle` calls like above (though less efficiently).
* **Rename**: Call attention to the implied coordinate system: `from_yxz_angles` (or is it `from_zxy_angles` ?)
* **Replace**: Replace with a more generic function, like `from_cardinal_angles(glam::Axes::YXZ, yaw, pitch, roll)`
* **Move**:  Hide it behind a trait, like `use glam::coordinate_systems::xyz_right_up_back::Rotation;`

bitshifter:
I did want to provide something at least and in this case at the time I wrote this I didn't want to try provide a generic solution. The main reason I wanted to provide something is a lot of beginners use Euler angles and I think struggle with other methods for creating rotations.

[...]

PR #163, opened 2021-04-12
add euler rotation sequence conversion

mickare:
Rust-port to convert quaternions to Euler rotation sequences.
Equations from: http://bediyap.com/programming/convert-quaternion-to-euler-rotations/

The two-axis Euler rotations (e.g. ZYZ) are not fully tested have to be treated with caution.
This is due how the test-implementation works and I haven't found a better generic way to test it.

### Usage
```rust
let q = Quat::from_euler(EulerRot::YXZi, 0.3, 0.1, 0.0);
let (y, x, z) = q.to_euler(EulerRot::YXZi);  // the "i" indicates intrinsic
```

```rust
let q = DQuat::from_euler(EulerRot::YXZi, 0.3, 0.1, 0.0);
let (y, x, z) = q.to_euler(EulerRot::YXZi);  // the "i" indicates intrinsic
```

### DIFF
--- a/src/core/traits/scalar.rs
+++ b/src/core/traits/scalar.rs
@@ -19,6 +19,7 @@ pub trait Signed: Sized + Num + core::ops::Neg<Output = Self> {
 
 #[cfg(not(feature = "libm"))]
 pub trait Float: Num + Copy + core::ops::Neg<Output = Self> {
+    fn asin(self) -> Self;
     fn acos(self) -> Self;
     fn ceil(self) -> Self;
     fn exp(self) -> Self;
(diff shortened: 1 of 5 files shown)


mickare:
The testing takes some time, to speed up things; the skipped tested angles could be increased in the test part at:
```rust
if u % 2 != 0 {
  continue;
}
```

bitshifter (commented):
Thanks for the PR. It goes some way to addressing #79. Would it be possible for the traits in `euler.rs` to be `pub(crate)`? It seems like they are only used internally so I'd rather not expose them to end users.

It might also be good to add a test to `tests/quat.rs` to make sure the euler version behaves the same as `Quat::from_rotation_ypr`.

mickare:
> Thanks for the PR. It goes some way to addressing #79. Would it be possible for the traits in `euler.rs` to be `pub(crate)`? It seems like they are only used internally so I'd rather not expose them to end users.

Yes, good idea, no need to expose them. c005157 and 0b56839 should address that.

> It might also be good to add a test to `tests/quat.rs` to make sure the euler version behaves the same as `Quat::from_rotation_ypr`.

There is one, but it is in the `euler.rs` test: cd6ef41

PR #164, opened 2021-04-14
Implement crevice traits

mtsr:
This implements crevice::AsStd140 for relevant types *except `BVec`*.

I wanted to check if something like this would be alright, before the remaining work:
* figuring out `BVec`
* adding the `AsStd430` trait
* anything else you think should be done.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,7 +40,7 @@ crevice_traits = ["crevice", "bytemuck"]
 
 [dependencies]
 bytemuck = { version = "1.5", optional = true, default-features = false }
-crevice = { version = "0.6", optional = true, path = "../crevice" }
+crevice = { version = "0.6", optional = true, git = "https://github.com/mtsr/crevice.git", branch = "develop" }
 mint = { version = "0.5", optional = true, default-features = false }
 num-traits = { version = "0.2.14", optional = true, default-features = false }
 rand = { version = "0.8", optional = true, default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,9 +35,12 @@ transform-types = []
 
 # libm is required when building no_std
 libm = ["num-traits/libm"]
+#
+crevice_traits = ["crevice", "bytemuck"]
 
 [dependencies]
 bytemuck = { version = "1.5", optional = true, default-features = false }
+crevice = { version = "0.6", optional = true, path = "../crevice" }
 mint = { version = "0.5", optional = true, default-features = false }
 num-traits = { version = "0.2.14", optional = true, default-features = false }
 rand = { version = "0.8", optional = true, default-features = false }
--- a/src/features/impl_crevice.rs
+++ b/src/features/impl_crevice.rs
@@ -1,0 +1,92 @@
+use crate::{
+    DMat2, DMat3, DMat4, DVec2, DVec3, DVec4, IVec2, IVec3, IVec4, Mat2, Mat3, Mat4, UVec2, UVec3,
+    UVec4, Vec2, Vec3, Vec4,
+};
+
+use bytemuck::Zeroable;
+
+macro_rules! vectors {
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -35,9 +35,12 @@ transform-types = []
 
 # libm is required when building no_std
 libm = ["num-traits/libm"]
+#
+crevice_traits = ["crevice", "bytemuck"]
 
 [dependencies]
 bytemuck = { version = "1.5", optional = true, default-features = false }
+crevice = { version = "0.6", optional = true, path = "../crevice" }
```
bitshifter, line 43: Need to specify the crates.io version here

Cargo.toml
```diff
@@ -35,9 +35,12 @@ transform-types = []
 
 # libm is required when building no_std
 libm = ["num-traits/libm"]
+#
+crevice_traits = ["crevice", "bytemuck"]
```
bitshifter, line 39: I'm not sure this is necessary? You should be able to check for crevice, you might need to add a bytemuck feature to the crevice dependency if it's not a default feature.

src/features/impl_crevice.rs
```diff
@@ -0,0 +1,92 @@
...
+macro_rules! matrices {
+            ( $( $glam_ty:ty, $std_name:ident, ( $($field:ident = $index:expr),* ), )* ) => {
+                $(
+                    impl crevice::std140::AsStd140 for $glam_ty {
+                        type Std140Type = crevice::std140::$std_name;
+
+                        fn as_std140(&self) -> Self::Std140Type {
+                            crevice::std140::$std_name {
+                                $(
+                                    $field: self.row($index).as_std140(),
```
bitshifter, line 62: Are GL matrices row major?

bitshifter (commented):
It looks OK although I don't know what crevice is doing behind the scenes :)

src/features/impl_crevice.rs
mtsr: Huh, that was in my first pass, I though I fixed that.

Cargo.toml
mtsr: I `use bytemuck::Zeroable;` in `impl_crevice.rs`, so I think I need the dep to enable crevice_traits?

bitshifter: I see, I thought crevice must have added bytemuck support via a feature, since they mention implementing bytemuck traits in their readme, but this doesn't seem to be the case.

Cargo.toml
```diff
@@ -35,9 +35,12 @@ transform-types = []
...
 
 [dependencies]
 bytemuck = { version = "1.5", optional = true, default-features = false }
+crevice = { version = "0.6", optional = true, git = "https://github.com/mtsr/crevice.git", branch = "develop" }
```
bitshifter, line 43: I wouldn't want to merge until this can use crates.io, I assume they need to make a new release or something?

Cargo.toml
CAD97: (just a drive-by following random threads as I do crate research)

If you want to "clever around" this and just allow deps to write `features = ["crevice"]`, you can:

- [rename the dependency](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html?highlight=rename,depende#renaming-dependencies-in-cargotoml) with e.g. `crevice_ = { package = "crevice", .. }`
- declare a `crevice = ["crevice_", "bytemuck"]` feature

It's a question of which hacky thing you want to do; rename the local crate name, or rename the feature.

-----

Alternatively, to eliminate the requirement on `bytemuck` for the impl you could

- Get `crevice` to implement `Default::default`, and use that (probably not a bad idea), or
- Name the padding fields in the macro expansion, and initialize the padding fields with `Default::default`.

bitshifter:
Going to close this as it's looking abandoned.

PR #170, opened 2021-04-30
Mat2 to simd

bitshifter:


bitshifter:
Removes some alignment and padding but performance is worse. Not going to do this for now.

PR #186, opened 2021-05-24
clamp input for acos_approx for f64

emilk:
I found that `DQuat::angle_between` would sometimes return `NaN` for small angles and tracked it down to this. There is likely a lot of other f64 angle-related function with similar problems.

### DIFF
--- a/src/core/traits/scalar.rs
+++ b/src/core/traits/scalar.rs
@@ -278,7 +278,7 @@ impl FloatEx for f64 {
     }
     #[inline(always)]
     fn acos_approx(self) -> Self {
-        f64::acos(self.clamp(-1.0, 1.0))
+        f64::acos(self.max(-1.0).min(1.0))
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/core/traits/scalar.rs
```diff
@@ -278,8 +278,7 @@ impl FloatEx for f64 {
     }
     #[inline(always)]
     fn acos_approx(self) -> Self {
-        // TODO: clamp range
```
bitshifter, line 281: I should look through my TODOs...

src/core/traits/scalar.rs
```diff
@@ -278,8 +278,7 @@ impl FloatEx for f64 {
...
-        f64::acos(self)
+        f64::acos(self.clamp(-1.0, 1.0))
```
bitshifter, line 281: `clamp` requires 1.50.0. Probably easier to use `self.max(-1.0).min(1.0)` instead for now.

bitshifter (approved):

bitshifter:
Thanks!

src/core/traits/scalar.rs
emilk: If you are using VS Code you might wanna try https://marketplace.visualstudio.com/items?itemName=Gruntfuggly.todo-tree :)

Issue #107, opened 2020-12-05
Add/Subtract Traits for vecs


DGriffin91:
It appears that add/subtract operator traits are only implemented for when lhs & rhs are the same type, but multiply/divide is implemented for more types.

For example, this works: `Vec3::one() * 3.0`
But this does not: `Vec3::one() + 3.0`

Is there something blocking this functionality, or some reason this shouldn't be implemented? I'm glad to help write implementations/tests if appropriate.

Thanks for all your work on this project!

bitshifter:
No real reason other than they don't really make sense mathematically, but as a programming convenience sure that could make sense. The only other negative I think is it might be a bit surprising, I try to be mostly explicit with glam. To be it would be slightly unexpected that you could add a scalar to a vector. Is this something your have found is supported in other libraries?

DGriffin91:
I see how mathematically it is odd. I've mostly run into the issue recently a bunch while converting shadertoy shaders to rust-gpu. 

GLSL supports it, so it gets used in GLSL shaders quite a bit. I come from a Python background and Python's Numpy also supports it.

I understand if it feels like it's not explicit enough and respect whatever decision is made. Since my backround has been with GLSL and Numpy I have assumed that this is a typical feature. I have since learned that it's not a given. (AFIK Ultraviolet & nalgebra do not support it)

I think adding a float to a vec is often done when the vec doesn't really represent a vector, or even a point.

PR #198, opened 2021-06-17
Implement element-wise Add and Sub for Vec types

SFBdragon:
It seems that while element wise operations are already implemented for multiply and divide operations on vector types, addition and subtraction are not. I'm unsure if this was a intended decision, however, I believe this is a simple improvement that increases convenience without sacrificing readability, and thus is a worthwhile feature.

The only concern I have is that this may be considered breaking in rare cases. See the modification of mat3.rs that was required to satisfy the compiler.

### DIFF
--- a/src/core/scalar/vector.rs
+++ b/src/core/scalar/vector.rs
@@ -202,6 +202,22 @@ impl<T: NumEx> Vector<T> for XY<T> {
     }
 
     #[inline]
+    fn add_scalar(self, other: T) -> Self {
+        Self {
+            x: self.x + other,
+            y: self.y + other,
+        }
+    }
+
+    #[inline]
+    fn sub_scalar(self, other: T) -> Self {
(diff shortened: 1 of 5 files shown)


bitshifter:
Hi there. This is something that has been requested in #107 , I've just been a bit on the fence about it. I'm inclined to accept the PR though since it seems like a reasonably common thing to expect (e.g. GLSL and numpy support this, probably other libs do as well).

SFBdragon:
Thank you for the merge, as well as the effort you have put into this library, it is much appreciated.

khyperia:
I guess just a misc comment/feedback from the void, when upgrading rust-gpu from glam 0.16 to 0.17 today, this did indeed break in various places, and so probably is a relatively commonly hit breaking change. Definitely okay to do, though, seems like a good change and is worth the break.

PR #223, opened 2021-08-05
Rkyv support

hannes-vernooij:
Added rkyv support

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,8 +40,8 @@ mint = { version = "0.5", optional = true, default-features = false }
 num-traits = { version = "0.2.14", optional = true, default-features = false }
 rand = { version = "0.8", optional = true, default-features = false }
 serde = { version = "1.0", optional = true, features = ["derive"] }
-rkyv = {version = "0.7", optional = true, default-features = false }
-bytecheck = { version = "0.6", optional = true, default-features = false }
+rkyv = { version = "0.7", optional = true }
+bytecheck = { version = "0.6", optional = true, default-features = false}
 
 [dev-dependencies]
 criterion = { version = "0.3", features = ["html_reports"] }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,9 +15,6 @@ maintenance = { status = "actively-developed" }
 
 [features]
 default = ["std"]
-rkyv_size_16 = ["rkyv/size_16"]
-rkyv_size_32 = ["rkyv/size_32"]
-rkyv_size_64 = ["rkyv/size_64"]
 
 # enable support for the standard library
 std = []
@@ -43,7 +40,7 @@ mint = { version = "0.5", optional = true, default-features = false }
 num-traits = { version = "0.2.14", optional = true, default-features = false }
 rand = { version = "0.8", optional = true, default-features = false }
 serde = { version = "1.0", optional = true, features = ["derive"] }
-rkyv = {version = "0.7.4", optional = true, default-features = false }
+rkyv = {version = "0.7", optional = true, default-features = false }
 bytecheck = { version = "0.6", optional = true, default-features = false }
 
 [dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,9 @@ maintenance = { status = "actively-developed" }
 
 [features]
 default = ["std"]
+rkyv_size_16 = ["rkyv/size_16"]
+rkyv_size_32 = ["rkyv/size_32"]
+rkyv_size_64 = ["rkyv/size_64"]
 
 # enable support for the standard library
 std = []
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -15,6 +15,9 @@ maintenance = { status = "actively-developed" }
 
 [features]
 default = ["std"]
+rkyv_size_16 = ["rkyv/size_16"]
+rkyv_size_32 = ["rkyv/size_32"]
+rkyv_size_64 = ["rkyv/size_64"]
```
djkoloski, line 20: These features are meant to be controlled by a top-level crate. I think that whatever uses glam and rkyv can choose these features and they'll unify all the way down.

Cargo.toml
```diff
@@ -40,6 +43,8 @@ mint = { version = "0.5", optional = true, default-features = false }
 num-traits = { version = "0.2.14", optional = true, default-features = false }
 rand = { version = "0.8", optional = true, default-features = false }
 serde = { version = "1.0", optional = true, features = ["derive"] }
+rkyv = {version = "0.7.4", optional = true, default-features = false }
```
djkoloski, line 46: I'd recommend changing this version dependency to `0.7`. The only changes going in 0.7.x will not be breaking.

src/features/impl_rkyv.rs
```diff
@@ -0,0 +1,197 @@
...
+            }
+        }
+    };
+
+    (@archive_deserialize $type:ty) => {
+        const _: () = {
+            #[cfg(not(any(feature = "archive_le", feature = "archive_be")))]
+            type Archived = $type;
+            #[cfg(feature = "archive_le")]
+            type Archived = rend::LittleEndian<$type>;
```
djkoloski, line 47: I'm not sure if this will work right. There might be some work to do here if you want glam types to support little- and big-endian, since some of the SIMD types are new primitives. If glam is using a standard crate I can see about getting this work done.

djkoloski (commented):
Thanks for getting the ball rolling on this! It looks really good and I'd like to help out where I can.

src/features/impl_rkyv.rs
hannes-vernooij: I removed the endian conversion step for now until support has been added to convert SIMD types.

djkoloski: Awesome, looks great! I'll be digging into this more comprehensively in https://github.com/rkyv/rend/issues/4

djkoloski (approved):
These all look good to me!

bitshifter:
Sorry I've been quite busy the last few days, I will look at this soon.

src/features/impl_rkyv.rs
```diff
@@ -0,0 +1,192 @@
...
+        impl<S: Fallible + ?Sized> Serialize<S> for $type {
+            #[inline]
+            fn serialize(&self, _: &mut S) -> Result<Self::Resolver, S::Error> {
+                Ok(())
+            }
+        }
+    };
+
+    (@archive_deserialize $type:ty) => {
+        const _: () = {
```
bitshifter, line 43: Why is this block necessary?

djkoloski: Looks like scoping for the type alias right below (`type Archived = $type`). Doesn't look strictly necessary to me.

bitshifter (approved):
Looks OK to me, just one question about the need for the `const` block.

src/features/impl_rkyv.rs
hannes-vernooij: Yes, it was to scope the type alias. I removed the scope.

bitshifter:
clippy failed on internal rkyv things, which is strange.

hannes-vernooij:
> clippy failed on internal rkyv things, which is strange.

If I understand the error correctly it means that size_16 size_32 or size_64 need to be passed as a feature by the crate using glam. I will look into getting size_32 enabled by default.

djkoloski:
This looks like it may be a situation where we want:

```toml
[dependencies]
rkyv = { version = "0.7", optional = true, default-features = false }

[dev-dependencies]
rkyv = "0.7"
```

However, this looks like it's only supported by the [feature resolver version 2](https://doc.rust-lang.org/cargo/reference/resolver.html#feature-resolver-version-2), which would bump the MSRV to 1.50. Feature resolver version 1 will incorrectly unify the default features from the dev-dependencies with the regular dependency features.

bitshifter:
I would be OK with switching to feature resolver version 2, it would address some other problems I hit in glam. That can happen separate to this PR though.

bitshifter:
@djkoloski incidentally it looks like rkyv's MSRV is 1.53 for use of or patterns (the readme says 1.52).

djkoloski:
Perfect, it's not an issue for rkyv users then. I'll probably follow suit and switch to feature resolver 2 (I believe there are some feature issues currently).

bitshifter:
Hmm, even with feature resolver 2 I'm still getting the clippy build fail if I disable default features on rkyv as suggested here https://github.com/bitshifter/glam-rs/pull/223#issuecomment-895328926

djkoloski:
Hmm, that's not the result I got when I did some testing. I'll take a look as soon as I have a chance.

PR #225, opened 2021-08-10
fix some warnings

remilauzier:


src/core/traits/scalar.rs
```diff
@@ -385,6 +385,6 @@ fn test_scalar_acos() {
     assert_approx_eq!((-2.0_f32).acos_approx(), core::f32::consts::PI);
 
     // input is clamped to -1.0..1.0
-    assert_eq!(2.0_f64.acos_approx(), 0.0);
-    assert_eq!((-2.0_f64).acos_approx(), core::f64::consts::PI);
+    assert!((2.0_f64.acos_approx() - 0.0).abs() < f64::EPSILON);
```
bitshifter, line 388: It's been a while since I wrote this but I think in this case I do want to check that the value is exactly equal, not within epsilon.

bitshifter (changes requested):

bitshifter:
Are these clippy warnings? It would be good to clarify, I haven't noticed them.

src/core/traits/scalar.rs
remilauzier: It seem clippy doesn't detect the case of being equal to zero so i will just revert that one and put a allow to silence it.

remilauzier: https://github.com/rust-lang/rust-clippy/issues/3804

CHANGELOG.md
```diff
@@ -332,7 +332,7 @@ The format is based on [Keep a Changelog], and this project adheres to
   `.y`, and `.z` instead of accessors.
 * `Quat`, `Vec3A` and `Vec4` implement `Deref` and `DerefMut` for the new `XYZ`
   and `XYZW` structs to emulate public named field access.
-* `Mat3` and `Mat4` had their axis members made pubic instead of needing
+* `Mat3` and `Mat4` had their axis members made public instead of needing
```
bitshifter, line 335: 🤣

bitshifter:
> ¨¨<'_>`` these are from rust_2018_idioms. Hidden anonymous lifetime are deprecated and wil be remove in rust 2021. the others are from clippy.

Ah right, it's a bit confusing because `deny(rust_2018_idioms)` sounds like it is disabling these warnings, but it is in-fact enabling these checks in clippy if I understand correctly.

remilauzier:
It's enabling them but it is in rustc. those of clippy always begin with ``clippy::``

src/lib.rs
```diff
@@ -235,6 +235,15 @@ The minimum supported Rust version is `1.45.0`.
...
 #![cfg_attr(target_arch = "spirv", feature(asm, register_attr, repr_simd))]
+#![deny(
+    rust_2018_compatibility,
+    rust_2018_idioms,
+    future_incompatible,
+    nonstandard_style
+)]
+// Wait until this bug is fix and float cmp to zero don't report a warning.
+// https://github.com/rust-lang/rust-clippy/issues/3804
+#![allow(clippy::float_cmp)]
```
bitshifter, line 246: Is there more than one of these generating warnings? If not it might be better to allow the specific line instead of allowing at the module level.

remilauzier: I will put it in the scalar.rs file since this is the only place where there is a wrong warning. But rust complain about not using an allow in a macro so i can't put in in the specific line.

src/core/traits/scalar.rs
```diff
@@ -386,5 +389,5 @@ fn test_scalar_acos() {
 
     // input is clamped to -1.0..1.0
     assert_eq!(2.0_f64.acos_approx(), 0.0);
-    assert_eq!((-2.0_f64).acos_approx(), core::f64::consts::PI);
+    assert!(((-2.0_f64).acos_approx() - core::f64::consts::PI).abs() < f64::EPSILON);
```
bitshifter, line 392: I just noticed that I have an approx comparison for the f32 implementation but that macro doesn't work for f64 (glam used to be f32 only, some things haven't gained f64 support).

src/lib.rs
bitshifter: Ah I see

src/core/traits/scalar.rs
remilauzier: Should i make some change?

bitshifter: No, just an aside.

bitshifter (approved):

bitshifter:
Thanks!

Issue #236, opened 2021-09-19
Primitive integer ops for UVec and IVec


bonsairobo:
I'm thinking about replacing my own vector types with glam, but there is some missing functionality I need first:

- Not
- BitAnd
- BitOr
- BitXor
- Shl
- Shr
- ~~Rem~~

Would a PR to add these `impl`s to `UVec` and `IVec` be welcome?

bitshifter:
Yep sure thing. I thought I added Rem already, maybe that was just on float types.

bonsairobo:
Ah yes. You `UVec` and `IVec` have `Rem` already.

bonsairobo:
I think I might need some guidance here.

My first take, based on reading the code, is that I should:

1. Add a `PrimInt` trait to bundle all of the scalar ops
2. Add an `IntegerVector` trait that bundles the vectorized ops
3. `impl IntegerVector` for `XY`, `XYZ`, and `XYZW`
4. `impl` the vectorized ops for `UVec` and `IVec` using a macro

Does that seem right, @bitshifter ?

bitshifter:
Ah yes, I have been meaning to document how things work internally for a while, but still haven't got around to it.

The way glam works is there are inner types, e.g. `XYZ` or SSE `__m128` and there are various traits that are implemented for those types. The public API is generally defined using macros and the implementation will be expressed using these inner types and trait implementations. Because the types you are looking at don't currently have a SIMD implementation you only need to worry about `XY`, `XYZ` and `XYZW`.

What you are proposing sounds about right. Perhaps `BitVector` as a trait name, `IntegerVector` sounds too broad to me. You may notice in code that I have  `Vector2`, `Vector3` and `Vector4` traits and so on. This is because these will all be implemented for SIMD types and for many ops you need to know how many elements you are dealing with. However I am not currently using SIMD for any of the `IVec*` or `UVec*` types so you can get away with a single trait. I think that's fine for now.

bonsairobo:
@bitshifter Any plans to support SIMD for `UVec` and `IVec`? I would be interested in that functionality.

bitshifter:
@bonsairobo I didn't have any plans to add SIMD support for these. SSE2 for one only supports signed integers in any case.

Also adding SIMD raises many questions like should there be an `IVec3A` type like there is for `f32`. It's a bit of a nuisance having two vector 3 types for `f32`.

I'd rather not do it for other primitive types unless there is a large benefit and strong demand from users.

[...]

PR #237, opened 2021-09-19
impl Not, Shl, Shr, BitAnd, BitOr, and BitXor for UVec and IVec

bonsairobo:
#236

### DIFF
--- a/src/core/scalar/vector.rs
+++ b/src/core/scalar/vector.rs
@@ -3,8 +3,6 @@ use crate::core::{
     traits::{scalar::*, vector::*},
 };
 
-use core::ops::{BitAnd, BitOr, BitXor, Not, Shl, Shr};
-
 impl<T: NumEx> VectorConst for XY<T> {
     const ZERO: Self = Self {
         x: <T as NumConstEx>::ZERO,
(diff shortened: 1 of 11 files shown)


src/core/scalar/vector.rs
```diff
@@ -1096,3 +1098,532 @@ impl SignedVector<f32> for XYZF32A16 {
 
 impl SignedVector3<f32> for XYZF32A16 {}
 impl FloatVector3<f32> for XYZF32A16 {}
+
+impl<T> Not for XY<T>
```
bitshifter, line 1102: I have not implemented traits directly on the inner types because I can't do the same thing for SIMD types due to the orphan rule. So my preference would be to provide a trait that has all the required methods like the existing traits work in the core traits module.

src/vec.rs
```diff
@@ -777,6 +777,109 @@ macro_rules! impl_vecn_signed_traits {
     };
 }
 
+macro_rules! impl_vecn_integer_traits {
```
bitshifter, line 780: I would prefer `impl_vecn_bitop_traits` or something as integer is too broad.

bitshifter (changes requested):
A couple of change requests.

Also would you consider adding some tests for all of these?

bonsairobo:
@bitshifter Yea happy to address all of your feedback and add tests.

bonsairobo:
I just learned some surprising stuff about shifting by negative numbers. Apparently, Rust won't let you shift by a negative number, since it would overflow. I assumed that it would just reverse the direction of the shift, but instead the negative RHS will get interpreted as unsigned (2s complement), and only the lower bits (within the LHS range) will get used by the shift operation.

So there is perhaps a judgement call to be made here: do we want to allow shifting by signed integers? I would side with whatever Rust does, which is to allow it, but it's only valid (at runtime) if the RHS is nonnegative.

bonsairobo:
The last patch makes all of the following possible in 2, 3, or 4 dimensions:

```
IVec << i8
IVec << i16
IVec << i32
IVec << u8
IVec << u16
IVec << u32

IVec << IVec
IVec << UVec

UVec << i8
UVec << i16
UVec << i32
UVec << u8
UVec << u16
UVec << u32

UVec << UVec
UVec << IVec
```

bonsairobo:
@bitshifter Any plans to support SIMD for `UVec` and `IVec`? I would be interested in that functionality.

bitshifter:
> 
> 
> I just learned some surprising stuff about shifting by negative numbers. Apparently, Rust won't let you shift by a negative number, since it would overflow. I assumed that it would just reverse the direction of the shift, but instead the negative RHS will get interpreted as unsigned (2s complement), and only the lower bits (within the LHS range) will get used by the shift operation.
> 
> So there is perhaps a judgement call to be made here: do we want to allow shifting by signed integers? I would side with whatever Rust does, which is to allow it, but it's only valid (at runtime) if the RHS is nonnegative.

I was kind of surprised that these ops were supported by signed integers!

Behaviour wise I tend to stick with whatever Rust does. As to whether shifting by a signed integer should be supported at all, I'm not sure. I tend to take a minimalist approach to features with glam. Generally if I don't use them and no one has asked for them (supported by a real use case) I don't tend to add things. So if you have no use for shifting by signed integers, or even different sizes (e.g. `i8`/`u8` etc.) then we could not add them for now.

bitshifter (approved):

bonsairobo:
I need at least the scalar ops for `u8` and `i32`. It wasn't much effort to add the rest. I'd like to keep them all for completeness.

bitshifter:

PR #238, opened 2021-09-27
Add `NAN` constants to all types

emilk:
`Vec2::NAN`, `Mat4::NAN` etc added which fills all values with `f32/f64::NAN`.

This is useful for marking a value as invalid (poisoning).

This is accomplished by having `f32`, `f64` and all private types implement this new trait:

``` rust
pub trait NanConstEx: Sized {
    const NAN: Self;
}
```


bitshifter:
Makes sense, thanks!

PR #244, opened 2021-10-04
Put wasm simd128 support on the glam-simd128 feature.

bitshifter:
Although the tests pass I spotted some bugs when enabling this in one of
my projects so this is on an experimental feature flag until those
issues are investigated and resolved.


bitshifter:
The simd128 bug I was seeing still happens if I disable the simd128 code path in glam, so I think the bug is probably something else. Since that is the main reason I made this change I think I'll not put this into main just yet.

PR #246, opened 2021-10-12
Implement mint conversions for Mat3A.

thomak-dev:
I am using Mat3A heavily in my project and I am dearly missing these.


bitshifter:
Thanks!

PR #255, opened 2021-11-13
Add "cuda" feature for enabling cuda-native alignment requirements on *Vec2 and *Vec4

anderslanglands:
CUDA prefers specific alignments for Vec2 and Vec4 types in order to use more efficient memory loads. This PR adds a "cuda" feature that enables those alignments using `cfg_attr` in order that glam can be used to generate efficient PTX code. 

This is currently a draft as the changes are somehow affecting DAffine2 and causing some of those tests to fail, where as far as I can see that type shouldn't be affected, but it's hard for me to trace back through all the macro invocations to understand where it's getting affected.

### DIFF
--- a/src/vec2.rs
+++ b/src/vec2.rs
@@ -181,8 +181,14 @@ macro_rules! impl_vec2_signed_traits {
 type XYF32 = XY<f32>;
 
 /// A 2-dimensional vector.
+#[cfg(not(feature = "cuda"))]
 #[derive(Clone, Copy)]
 #[repr(transparent)]
+pub struct Vec2(pub(crate) XYF32);
+
+#[cfg(feature = "cuda")]
+#[derive(Clone, Copy)]
+#[repr(C, align(8))]
 pub struct Vec2(pub(crate) XYF32);
 
 impl Vec2 {
@@ -195,9 +201,16 @@ impl_vec2_signed_traits!(f32, vec2, Vec2, Vec3, BVec2, XYF32);
 
 type XYF64 = XY<f64>;
 
+#[cfg(not(feature = "cuda"))]
 /// A 2-dimensional vector.
 #[derive(Clone, Copy)]
 #[repr(transparent)]
+pub struct DVec2(pub(crate) XYF64);
+
+#[cfg(feature = "cuda")]
+/// A 2-dimensional vector.
+#[derive(Clone, Copy)]
+#[repr(C, align(16))]
 pub struct DVec2(pub(crate) XYF64);
 
 impl DVec2 {
@@ -210,9 +223,16 @@ impl_vec2_signed_traits!(f64, dvec2, DVec2, DVec3, BVec2, XYF64);
 
 type XYI32 = XY<i32>;
 
+#[cfg(not(feature = "cuda"))]
 /// A 2-dimensional vector.
 #[derive(Clone, Copy)]
 #[repr(transparent)]
+pub struct IVec2(pub(crate) XYI32);
+
+#[cfg(feature = "cuda")]
+/// A 2-dimensional vector.
+#[derive(Clone, Copy)]
+#[repr(C, align(8))]
 pub struct IVec2(pub(crate) XYI32);
 
 impl IVec2 {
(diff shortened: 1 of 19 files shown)


bitshifter:
The reason it is failing is due to the intersection of `cuda` and `scalar-math` features. The `scalar-math` feature disables SIMD and thus the SIMD alignment. Ideally your changes should work with `scalar-math` as well, which will involve a bit more work.

The reason it's failing in the lint step is because clippy is run with `--all-feature` which will enable both `cuda` and `scalar-math`.

You should also add the following to `FEATURE_SETS` in `build_and_test_features`:

```sh
"std cuda"
"std scalar-math cuda"
```

It shouldn't be necessary to test `cuda` with any other glam features, just `std` and with and without `scalar-math`.

src/vec2.rs
```diff
@@ -268,6 +272,7 @@ impl_vecn_scalar_bit_op_traits!(UVec2, u32, XYU32);
 
 impl_vecn_bit_op_traits!(UVec2, XYU32);
 
+#[cfg(not(feature = "cuda"))]
```
bitshifter, line 275: I think it would be preferable to use `#[cfg]` blocks inside a common test instead of duplicating the tests. Primarily because which alignment is different, size should be the same. It looks like you've taken this approach in the `tests/*.rs` changes already.

bitshifter: On second thought I see this is following the style of existing code, so ignore me. I'll make that change down the line.

bitshifter:
The compile fail is easily fixed: 

```diff
@@ -491,7 +491,10 @@ impl_affine2_traits!(
     DerefTargetF64
 );
 
-#[cfg(any(feature = "scalar-math", target_arch = "spriv"))]
+#[cfg(all(
+    not(feature = "cuda"),
+    any(feature = "scalar-math", target_arch = "spriv")
+))]
 mod const_test_affine2 {
     const_assert_eq!(
         core::mem::align_of::<f32>(),
```

After fixing that there is a test failure with `--features=cuda`:

```
failures:

---- daffine2::test_affine2_fmt stdout ----
thread 'daffine2::test_affine2_fmt' panicked at 'assertion failed: `(left == right)`
  left: `"[[5, 6], [1, 2], [3, 4]]"`,
 right: `"[[1, 2], [3, 4], [5, 6]]"`', tests/affine2.rs:213:5
```

Which is quite strange. I suspect it's to do with `Deref`, but I would have thought making alignment `8` would not change layout of any glam types, so I'm not sure what's going on there.

bitshifter:
The problem is that `Mat2` and `DMat2` are not aligned the same as `Vec2` and `DVec2`, this is because matrices are composed of inner types which aren't `cuda` aligned. In some places `Mat2` and `DMat2` use `Deref` with a target of `Colums2<Vec2>` or `Columns2<DVec2>` but these are not following the same alignment rules as `Vec2` and `DVec2` which results in undefined behaviour. It can be fixed, with the right combination of `#[cfg]` blocks, I already deal with this problem for `scalar-math`, but it will be messy.

The reason why glam does all this stuff is explained (hopefully) in https://github.com/bitshifter/glam-rs/blob/main/ARCHITECTURE.md.

bitshifter:
This made me realize that the affine types should be `repr(C)`, I've fixed that on main now. I think that would be the reason the `fmt` test was failing. I still get a segfault in mat2 tests which I think is most likely due to alignment mismatching.

Also, I made this PR non draft by accident, pressed the wrong button on my phone :)

anderslanglands:
Yeah that's the part I don't understand - I'm only defining the cuda
alignments on the Vec* types, while Mat2 uses Column2<XY<T>> which
shouldn't be affected by that?

anderslanglands:
ahhh because it's also implementing deref to Column2<Vec2>. Hmmm.

anderslanglands:
Simplest solution might just be to align Mat2/DMat2 to match the relevant
Vec type... would that work?

bitshifter:
The tests currently fail on `cargo clippy --all-features`, but it's really just `cargo test --features cuda,scalar-math`

anderslanglands:
Think that should do it now.

bitshifter (approved):

bitshifter:
Nice  work on pushing this through. I might get rid of `repr(transparent)` which would make all these cfg blocks a lot simpler. I doubt I'm getting any benefits from transparent.

anderslanglands:
Yeah probably not. Thanks for the merge!

PR #262, opened 2021-12-11
Fix typo: `tha` -> `than`

hatoo:
I believe I found a typo.
Sorry if it's intentional.

Thanks for creating `glam`!


bitshifter:
Thanks!

PR #263, opened 2021-12-17
Replace spriv typo with spirv

khyperia:
I've made this typo _so_ many many countless times myself, haha.

Noticed this when upgrading to glam 0.20 - `Vec4` wasn't marked as `repr(transparent)` due to the typo, which unfortunately mucked up a cast in Mat4's Deref impl from `*const Columns4<XYZW<f32>>` to `*const Columns4<Vec4>` in rust-gpu (due to the extra layer of struct wrapper in the latter)... it'd be real nice if SPIR-V was more friendly and allowed casts like that, but alas. I've confirmed this change fixes rust-gpu's consumption of glam.


bitshifter:
Ah whoops, looks like I've been very consistent misspelling that 😅

PR #267, opened 2021-12-27
Add `Quat::to_array` function

rafalh:
Add `Quat::to_array` function to simplify conversion from `Quat` type to plain array.
`Quat` already has `from_array` function and vector types have `from_array` function so as a user of this library I would expect to have such function implemented for `Quat` struct as well.

### DIFF
--- a/src/quat.rs
+++ b/src/quat.rs
@@ -282,6 +282,12 @@ macro_rules! impl_quat_methods {
             euler.convert_quat(self)
         }
 
+        /// `[x, y, z, w]`
+        #[inline(always)]
+        pub fn to_array(&self) -> [$t; 4] {
+            [self.x, self.y, self.z, self.w]
+        }
+
         /// Returns the vector part of the quaternion.
         #[inline(always)]
         pub fn xyz(self) -> $vec3 {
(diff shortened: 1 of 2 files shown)


bitshifter:
Could you add a test please.

eadf:
my 2 cents:
Isn't this functionally the same as the already existing `<[f32;4]>::from(quat)`?

Sidenote: The reverse `From` is not implemented so maybe `Quat::from([f32;4])` should be implemented too?
```rust
/// Implements Quat::from([f32;4]);
/// ```
/// use glam::Quat;
/// let a = [0.1_f32,0.2,0.3,0.4];
/// let q = Quat::from(a);
/// assert_eq!(a[0],q.x);
/// assert_eq!(a[1],q.y);
/// assert_eq!(a[2],q.z);
/// assert_eq!(a[3],q.w);
/// ```
impl From<[f32;4]> for Quat {
    fn from(array: [f32; 4]) -> Self {
        Quat::from_array(array)
    }
}
```
But in a macro that covers all variants.

bitshifter:
I removed `<[f32;4]`::from(Quat)` because I felt it would be too easy to
create a non uniform quat without realising, especially because docs for
trait implementations are a bit hidden.

The `to_array` and `from_array` methods on vectors etc is in part to mirror
the methods on matrices, matrices however specify row or column ordering of
the array, which isn't necessary for the vector types.

PR #270, opened 2021-12-30
Expose mul_add for vector types.

### LINKED ISSUE #269: Expose FMA via "mul_add" for f32/f64 Vec types.
This should allow users to ensure that a FMA instruction is used (with the SSE2 instruction if possible), which should both improve the accuracy and potentially speed of the resultant code.

james7132:
Fixes #269.

This PR does the following:

 * Exposes "mul_add" on all vector types.
 * Updates documentation for the function
 * Forces the scalar implementations to use `f32::mul_add` and `f64::mul_add` for the types where it's relevant.
 * Updates the relevant test names as "mul_add" is no longer private.


eadf:
This looks like infinite recursion when it actually isn't:
```rust
fn mul_add(self, b: Self, c: Self) -> Self { self.mul_add(b, c) }
```

Maybe we could avoid confusion and bring it in line with how `min()` & `max()` are implemented by writing it as:
```rust
fn mul_add(self, b: Self, c: Self) -> Self { f64::mul_add(self, b, c) }
```
(same thing with f32)

bitshifter:
It would be good to see some comparison benchmarks for this. I think `mul_add` is used internally in some places and I would want to be confident this doesn't regress performance.

The reason why I have this concern is ultraviolet used to use FMA heavily but decided it was ultimately not worthwhile and removed it, see https://github.com/termhn/ultraviolet/issues/8#issuecomment-805315569

So if there was a regression from this change with regards to internal use I could still accept this if internal use was changed to not use `mul_add`, then it's just a public function that people can decide to use or not, but the comments should probably include the caveats about using FMA.

james7132:
Sorry for the delay in this. I still haven't had the time to run benchmarks, I'll get to that soon.

With this said, even without the benchmarks, I recently realized that the fact that `m128_mul_add` is used for Vec3A/Vec4 on targets with SSE2 support, but two separate operations are used in WASM and the scalar implementations means that the resultant output will be different on different targets (with or without SSE2 support), or different results for Vec3 vs Vec3A on SSE2 targets. This typically isn't an issue, but this represents a determinism issue for systems that are reliant on Rust (and by proxy glam) to provide identical floating point outputs.

Could we have both implementations and just switch based on a `enhanced-determinism` flag or something similar?

alice-i-cecile:
> Could we have both implementations and just switch based on a enhanced-determinism flag or something similar?

As someone with pretty serious interest in using Bevy (and hence glam) for scientific computing, an `enhanced-determinism` flag would be incredibly valuable (if we can't get determinism by default).

bitshifter:
I think it would be best if this kind of thing was opt-in rather than opt-out. glam tries to be deterministic as possible out of the box (e.g. avoiding problematic instructions like `_mm_rsqrt_ps` and `_mm_rcp_ps` on SSE2.

james7132:
If that's the case, I think it's best to actually just opt for using separate multiply/add operations within each of them, since WASM has no `mul_add` support outside of the scalar `mul_add` implementations, then have a `break-determinism` feature flag that enables `m128_mul_add` on the SSE2 types. This would avoid making syscalls on the scalar implementations at the cost of one more instruction for SIMD-based targets and obviously the accuracy cost of one more rounding error.

Alternatively, we could use the fused multiply add instructions by default and resort to a scalar split for WASM, but that can be quite a negative impact on perf by default for the other platforms.

james7132:
On second thought, feature flags are effectively `OR`'ed across the dependency tree. This presents a usability issue where crates that use a `break-determinism` flag may break the invariant for other crates that don't. While some crates can live with degraded perf from additional determinism, those reliant on deterministic operations will have their base assumptions broken. Perhaps opt-out might actually be the way here.

bitshifter:
AFAIK `fma` is not enabled by default in `x86` or `x86_64` targets, so in that sense it needs to be opt in anyway. I don't really have an issue with making `mul_add` public and providing an SSE implementation if the `fma` feature is enabled. It's really up to users if they want to enable that. Perhaps that gets more complicated when support is added for an instruction set where FMA is enabled by default, at that point perhaps another flag may be required.

It would be good to benchmark before bike shedding too much I think.

james7132:
Did a full benchmark of `main`, the current head of this branch (`no-sse-mul-add`) for determinism, and the earlier full `fma` implementation

```
group                          full-fma                               main                                   no-sse-mul-add
-----                          --------                               ----                                   --------------
affine2 from srt               1.00     11.2±0.37ns        ? ?/sec    1.10     12.3±0.62ns        ? ?/sec    1.02     11.4±0.86ns        ? ?/sec
affine2 inverse                1.24      2.9±0.15ns        ? ?/sec    1.26      2.9±0.17ns        ? ?/sec    1.00      2.3±0.05ns        ? ?/sec
affine2 mul affine2            1.00      1.8±0.10ns        ? ?/sec    1.01      1.9±0.18ns        ? ?/sec    1.00      1.8±0.07ns        ? ?/sec
affine2 mul mat3               1.06      5.4±0.44ns        ? ?/sec    1.00      5.1±0.10ns        ? ?/sec    1.07      5.4±0.09ns        ? ?/sec
affine2 transform point2       1.00      1.2±0.04ns        ? ?/sec    1.06      1.3±0.12ns        ? ?/sec    1.01      1.2±0.03ns        ? ?/sec
affine2 transform vector2      1.00      1.1±0.05ns        ? ?/sec    1.52      1.7±0.49ns        ? ?/sec    1.05      1.2±0.05ns        ? ?/sec
affine3a from srt              1.00      4.4±0.07ns        ? ?/sec    1.01      4.4±0.08ns        ? ?/sec    1.01      4.4±0.17ns        ? ?/sec
affine3a inverse               1.00      6.3±0.57ns        ? ?/sec    1.02      6.4±0.15ns        ? ?/sec    1.02      6.4±0.16ns        ? ?/sec
affine3a mul affine3a          1.14      4.1±0.49ns        ? ?/sec    1.02      3.7±0.25ns        ? ?/sec    1.00      3.6±0.17ns        ? ?/sec
affine3a mul mat4              1.01      5.4±0.10ns        ? ?/sec    1.00      5.3±0.08ns        ? ?/sec    1.01      5.4±0.14ns        ? ?/sec
affine3a transform point3      1.14      1.9±0.08ns        ? ?/sec    1.00      1.7±0.05ns        ? ?/sec    1.07      1.8±0.13ns        ? ?/sec
affine3a transform point3a     1.14      1.6±0.09ns        ? ?/sec    1.00      1.4±0.07ns        ? ?/sec    1.00      1.4±0.05ns        ? ?/sec
affine3a transform vector3     1.18      1.9±0.08ns        ? ?/sec    1.01      1.6±0.04ns        ? ?/sec    1.00      1.6±0.03ns        ? ?/sec
affine3a transform vector3a    1.08      1.5±0.11ns        ? ?/sec    1.00      1.3±0.04ns        ? ?/sec    1.00      1.3±0.02ns        ? ?/sec
mat2 determinant               1.00      0.8±0.04ns        ? ?/sec    1.00      0.8±0.04ns        ? ?/sec    1.03      0.9±0.05ns        ? ?/sec
mat2 inverse                   1.03      1.3±0.05ns        ? ?/sec    1.00      1.3±0.07ns        ? ?/sec    1.00      1.2±0.02ns        ? ?/sec
mat2 mul mat2                  1.00      1.2±0.27ns        ? ?/sec    1.29      1.5±0.49ns        ? ?/sec    1.14      1.3±0.62ns        ? ?/sec
mat2 mul vec2                  1.05      1.1±0.07ns        ? ?/sec    1.03      1.1±0.06ns        ? ?/sec    1.00      1.1±0.02ns        ? ?/sec
mat2 transpose                 1.23      1.3±0.43ns        ? ?/sec    1.00      1.1±0.21ns        ? ?/sec    1.22      1.3±0.40ns        ? ?/sec
mat3 determinant               1.14      2.0±0.13ns        ? ?/sec    1.02      1.8±0.14ns        ? ?/sec    1.00      1.8±0.03ns        ? ?/sec
mat3 from ypr                  1.01     53.4±2.67ns        ? ?/sec    1.00     52.8±1.69ns        ? ?/sec    1.01     53.3±3.57ns        ? ?/sec
mat3 inverse                   1.08      8.2±0.65ns        ? ?/sec    1.04      7.9±0.37ns        ? ?/sec    1.00      7.6±0.52ns        ? ?/sec
mat3 mul affine2               1.03      5.0±0.35ns        ? ?/sec    1.01      4.8±0.30ns        ? ?/sec    1.00      4.8±0.06ns        ? ?/sec
mat3 mul mat3                  1.00      5.3±0.06ns        ? ?/sec    1.08      5.7±0.44ns        ? ?/sec    1.01      5.4±0.15ns        ? ?/sec
mat3 mul vec3                  1.00      2.5±0.03ns        ? ?/sec    1.00      2.5±0.03ns        ? ?/sec    1.08      2.7±0.11ns        ? ?/sec
mat3 mul vec3a                 1.00      2.0±0.03ns        ? ?/sec    1.01      2.0±0.09ns        ? ?/sec    1.06      2.1±0.03ns        ? ?/sec
mat3 transform point2          1.00      1.4±0.03ns        ? ?/sec    1.00      1.4±0.03ns        ? ?/sec    1.01      1.4±0.04ns        ? ?/sec
mat3 transform vector2         1.00      1.2±0.05ns        ? ?/sec    1.03      1.3±0.05ns        ? ?/sec    1.03      1.3±0.08ns        ? ?/sec
mat3 transpose                 1.00      2.1±0.06ns        ? ?/sec    1.00      2.1±0.02ns        ? ?/sec    1.01      2.1±0.03ns        ? ?/sec
mat3a determinant              1.00      1.4±0.02ns        ? ?/sec    1.01      1.4±0.07ns        ? ?/sec    1.02      1.4±0.03ns        ? ?/sec
mat3a from ypr                 1.09     54.6±3.48ns        ? ?/sec    1.01     50.2±1.20ns        ? ?/sec    1.00     49.9±0.68ns        ? ?/sec
mat3a inverse                  1.01      4.3±0.22ns        ? ?/sec    1.01      4.4±0.13ns        ? ?/sec    1.00      4.3±0.07ns        ? ?/sec
mat3a mul mat3a                1.05      2.7±0.28ns        ? ?/sec    1.00      2.6±0.09ns        ? ?/sec    1.08      2.8±0.12ns        ? ?/sec
mat3a mul vec3                 1.01      1.6±0.03ns        ? ?/sec    1.01      1.6±0.03ns        ? ?/sec    1.00      1.6±0.03ns        ? ?/sec
mat3a mul vec3a                1.00      1.3±0.04ns        ? ?/sec    1.03      1.3±0.10ns        ? ?/sec    1.01      1.3±0.09ns        ? ?/sec
mat3a transform point2         1.00      1.3±0.12ns        ? ?/sec    1.01      1.3±0.15ns        ? ?/sec    1.03      1.4±0.08ns        ? ?/sec
mat3a transform vector2        1.06      1.3±0.11ns        ? ?/sec    1.09      1.4±0.12ns        ? ?/sec    1.00      1.2±0.02ns        ? ?/sec
mat3a transpose                1.04      1.5±0.09ns        ? ?/sec    1.03      1.4±0.05ns        ? ?/sec    1.00      1.4±0.02ns        ? ?/sec
mat4 determinant               1.00      4.1±0.07ns        ? ?/sec    1.00      4.1±0.10ns        ? ?/sec    1.00      4.1±0.11ns        ? ?/sec
mat4 from srt                  1.00      4.9±0.17ns        ? ?/sec    1.00      4.9±0.15ns        ? ?/sec    1.04      5.1±0.21ns        ? ?/sec
mat4 from ypr                  1.00     51.1±0.62ns        ? ?/sec    1.10     56.3±3.92ns        ? ?/sec    1.00     51.3±0.82ns        ? ?/sec
mat4 inverse                   1.03     14.0±0.83ns        ? ?/sec    1.03     13.9±0.97ns        ? ?/sec    1.00     13.5±1.09ns        ? ?/sec
mat4 mul affine3a              1.00      3.8±0.05ns        ? ?/sec    1.00      3.8±0.07ns        ? ?/sec    1.04      3.9±0.30ns        ? ?/sec
mat4 mul mat4                  1.02      4.6±0.14ns        ? ?/sec    1.00      4.5±0.08ns        ? ?/sec    1.01      4.6±0.13ns        ? ?/sec
mat4 mul vec4                  1.09      1.6±0.12ns        ? ?/sec    1.00      1.5±0.02ns        ? ?/sec    1.03      1.5±0.06ns        ? ?/sec
mat4 transform point3          1.02      1.7±0.04ns        ? ?/sec    1.00      1.7±0.03ns        ? ?/sec    1.01      1.7±0.12ns        ? ?/sec
mat4 transform point3a         1.14      1.6±0.08ns        ? ?/sec    1.00      1.4±0.06ns        ? ?/sec    1.01      1.4±0.12ns        ? ?/sec
mat4 transform vector3         1.03      1.7±0.05ns        ? ?/sec    1.00      1.6±0.02ns        ? ?/sec    1.03      1.7±0.13ns        ? ?/sec
mat4 transform vector3a        1.00      1.4±0.11ns        ? ?/sec    1.00      1.4±0.04ns        ? ?/sec    1.03      1.5±0.08ns        ? ?/sec
mat4 transpose                 1.02      1.8±0.03ns        ? ?/sec    1.00      1.8±0.07ns        ? ?/sec    1.02      1.8±0.11ns        ? ?/sec
quat conjugate                 1.15      1.1±0.06ns        ? ?/sec    1.11      1.1±0.18ns        ? ?/sec    1.00      1.0±0.09ns        ? ?/sec
quat dot                       1.00      1.0±0.03ns        ? ?/sec    1.01      1.0±0.02ns        ? ?/sec    1.09      1.1±0.07ns        ? ?/sec
quat from ypr                  1.00     42.2±0.56ns        ? ?/sec    1.00     42.2±0.73ns        ? ?/sec    1.14     48.1±2.06ns        ? ?/sec
quat lerp                      1.13      4.6±0.37ns        ? ?/sec    1.00      4.1±0.08ns        ? ?/sec    1.00      4.1±0.13ns        ? ?/sec
quat mul quat                  1.07      1.9±0.14ns        ? ?/sec    1.02      1.9±0.12ns        ? ?/sec    1.00      1.8±0.03ns        ? ?/sec
quat mul vec3                  1.00      3.7±0.11ns        ? ?/sec    1.12      4.2±0.19ns        ? ?/sec    1.00      3.7±0.10ns        ? ?/sec
quat mul vec3a                 1.00      3.4±0.04ns        ? ?/sec    1.13      3.8±0.25ns        ? ?/sec    1.10      3.7±0.03ns        ? ?/sec
quat slerp                     1.05     19.1±1.13ns        ? ?/sec    1.03     18.7±0.61ns        ? ?/sec    1.00     18.2±0.26ns        ? ?/sec
vec2 angle_between             1.00      9.3±0.08ns        ? ?/sec    1.11     10.3±0.52ns        ? ?/sec    1.00      9.3±0.39ns        ? ?/sec
vec2 euler                     1.00     10.9±0.13µs        ? ?/sec    1.00     10.9±0.32µs        ? ?/sec    1.01     11.0±0.42µs        ? ?/sec
vec2 mul vec2                  1.00      1.7±0.04ns        ? ?/sec    1.01      1.7±0.03ns        ? ?/sec    1.01      1.7±0.07ns        ? ?/sec
vec2 select                    1.00      1.3±0.04ns        ? ?/sec    1.00      1.3±0.05ns        ? ?/sec    1.02      1.4±0.07ns        ? ?/sec
vec3 angle_between             1.00      9.4±0.20ns        ? ?/sec    1.05      9.9±0.27ns        ? ?/sec    1.09     10.3±0.30ns        ? ?/sec
vec3 any_orthogonal_vector     1.58      2.5±0.79ns        ? ?/sec    1.00      1.6±0.65ns        ? ?/sec    1.34      2.1±0.68ns        ? ?/sec
vec3 any_orthonormal_pair      1.07      2.7±0.25ns        ? ?/sec    1.00      2.5±0.05ns        ? ?/sec    1.00      2.5±0.18ns        ? ?/sec
vec3 any_orthonormal_vector    1.08      1.9±0.11ns        ? ?/sec    1.01      1.8±0.10ns        ? ?/sec    1.00      1.8±0.04ns        ? ?/sec
vec3 euler                     1.02     21.1±1.62µs        ? ?/sec    1.00     20.7±0.72µs        ? ?/sec    1.07     22.1±1.29µs        ? ?/sec
vec3 into array fast           1.07      1.0±0.07ns        ? ?/sec    1.02      1.0±0.01ns        ? ?/sec    1.00      0.9±0.02ns        ? ?/sec
vec3 into array fields         1.00      0.9±0.01ns        ? ?/sec    1.09      1.0±0.06ns        ? ?/sec    1.12      1.1±0.06ns        ? ?/sec
vec3 into tuple fast           1.00      0.9±0.01ns        ? ?/sec    1.05      1.0±0.03ns        ? ?/sec    1.04      1.0±0.03ns        ? ?/sec
vec3 mul vec3                  1.13      2.1±0.16ns        ? ?/sec    1.17      2.2±0.04ns        ? ?/sec    1.00      1.9±0.03ns        ? ?/sec
vec3 normalize                 1.00      2.7±0.08ns        ? ?/sec    1.07      2.9±0.11ns        ? ?/sec    1.03      2.8±0.09ns        ? ?/sec
vec3 normalize_or_zero         1.09      4.6±0.16ns        ? ?/sec    1.00      4.3±0.34ns        ? ?/sec    1.01      4.3±0.15ns        ? ?/sec
vec3 select                    1.00      1.8±0.02ns        ? ?/sec    1.02      1.9±0.03ns        ? ?/sec    1.00      1.8±0.02ns        ? ?/sec
vec3 to rgb                    1.00      2.6±0.06ns        ? ?/sec    1.06      2.8±0.22ns        ? ?/sec    1.06      2.8±0.18ns        ? ?/sec
vec3a angle_between            1.00      8.9±0.05ns        ? ?/sec    1.00      8.9±0.11ns        ? ?/sec    1.01      9.0±0.20ns        ? ?/sec
vec3a euler                    1.07      6.9±0.11µs        ? ?/sec    1.01      6.5±0.33µs        ? ?/sec    1.00      6.4±0.06µs        ? ?/sec
vec3a into array deref         1.10      1.1±0.08ns        ? ?/sec    1.00      1.0±0.00ns        ? ?/sec    1.11      1.1±0.01ns        ? ?/sec
vec3a into array fast          1.00      0.9±0.00ns        ? ?/sec    1.02      0.9±0.03ns        ? ?/sec    1.05      0.9±0.00ns        ? ?/sec
vec3a into tuple fast          1.00      0.9±0.02ns        ? ?/sec    1.01      0.9±0.01ns        ? ?/sec    1.02      0.9±0.03ns        ? ?/sec
vec3a into vec3                1.00      0.9±0.01ns        ? ?/sec    1.20      1.1±0.03ns        ? ?/sec    1.04      1.0±0.03ns        ? ?/sec
vec3a mul vec3a                1.00      1.2±0.01ns        ? ?/sec    1.05      1.3±0.04ns        ? ?/sec    1.01      1.2±0.05ns        ? ?/sec
vec3a select                   1.00      1.3±0.02ns        ? ?/sec    1.01      1.3±0.03ns        ? ?/sec    1.04      1.3±0.04ns        ? ?/sec
vec3a to rgb                   1.00      2.7±0.04ns        ? ?/sec    1.01      2.7±0.15ns        ? ?/sec    1.03      2.8±0.22ns        ? ?/sec
vec4 mul vec4                  1.15      1.2±0.01ns        ? ?/sec    1.20      1.2±0.01ns        ? ?/sec    1.00      1.0±0.02ns        ? ?/sec
vec4 select                    1.00      1.3±0.02ns        ? ?/sec    1.10      1.4±0.02ns        ? ?/sec    1.07      1.4±0.04ns        ? ?/sec
```

bitshifter:
> I might be biased, but I think the `no-sse-mul-add` option offers both the best results here and cross-platform determinism. There are a few odd and egregious cases when comparing the results from main and the other two, and I'm not sure if that's just bad sampling or the difference is genuinely that bad.

To clarify, that is using `f32::mul_add` and `f64::mul_add` and avoiding the fma sse2 intrinsics? Was it built with `-C target-feature+=fma`?

Results are a bit all over the place, things like `Mat2::transpose()` shouldn't really be affected by fma at all, but is slower here for some reason.

What might be more interesting is if you have some code that can make use of the public `mul_add` implementation and see if it is sped up. Probably glam's internal use of `mul_add` isn't all that significant and might not benefit particularly.

Feature wise I'm not sure about the `break-determinism` name. Float determinism is not all that clear cut. For example, fmamight be considered deterministic enough if you are using fixed hardware. E.g. some of the instructions I avoid are not deterministic between AMD and Intel chips, but if you were only targeting AMD (e.g. like a console) that wouldn't matter. Traditionally in C++ these kind of optimizations are enabled by `fast-math` in gcc/clang/msvc, so perhaps that would be a more appropriate name? It could also enable some of the `rcp` and `rsqrt` as well at the expense of determinism and accuracy in the `rcp`, `rsqrt` case.

james7132:
No that particular benchmark is basically `main` without the SSE2 FMA.

My original main motivation (before the determinism concern) for this was to speed up particle simulation for particle systems in Bevy, hoping to use the SSE FMA on Vec4, but the determinism concern is a bigger issue for me. I'll see if I can make a benchmark.

FMA is standardized in IEEE-754 2008, which is what both LLVM and Rust's f32/f64 types (claim to) implement. Other operations like the trig functions and rsqrt aren't, so I'm tentatively relying on libm to provide consistent and deterministic implementations there in a cross platform way.

bitshifter:
Yeah it does look like `native` enables `fma` on `x86_64` https://rust.godbolt.org/z/qrKbdedMf.

I wasn't sure if fma was optional, it seems it is required https://irem.univ-reunion.fr/IMG/pdf/ieee-754-2008.pdf.

I do wonder what happens if there is no hardware support though, if there is some kind of software emulation that is used instead (e.g. when using `f32::mul_add`).

Tried that in godbolt, if `fma` is enabled then it uses the appropriate sse instruction, otherwise it's a jump to something that I assume would be software emulation, but I can't see that in godbolt. https://rust.godbolt.org/z/41WvK3Pv8

I think it's a tricky thing to know what good behaviour would be. If we provide `mul_add` is should actually perform a fused multiply add, but that might actually be slower than separate multiply and add if there is no hardware support.

My inclination would be to make the public `mul_add` always do a fused multiply add using (`f32::mul_add`, `f64::mul_add` or sse intrinsics if available) but docs should warn that this could be slower if there is no hardware support.

For glam internals I would lean towards not using fused multiply add unless `fast-math` is enabled and then only do it if there is hardware support (and warn that this could give different results on different architectures).

bitshifter:
I stepped through the disassembly for the software fmaf that is used if hardware support is enabled (this is msvc toolchain on x86_64). The disassembly is here https://gist.github.com/bitshifter/b53df0274bd16c15f7722501afdbc10b

It involves a function call and it appears to be doing a runtime check for hardware support and using it if it's available (which it is for me on my x86_64 machine), but I think this is going to be pretty expensive compared to doing a separate mul and add if hardware support is not enabled.

So, I think this reinforces my previous comment:

* I'm happy to make `VecN::mul_add` public and have it use `f32::mul_add`, `f64::mul_add` or `_mm_madd_ps`, the comment should mention that this will produce different results to a separate mul and add and that without hardware support enabled it will most likely be slower. I don't think this needs to be on a `fast-math` flag.
* Internally I would only use `mul_add` if `fast-math` feature is enable AND hardware support is enabled, so might need an internal implementation that has those config checks and does the right thing

james7132:
@bitshifter done! I must admit that unpacking individual `f32`s out of a SIMD vector to individually call `mul_add` on them is difficult pill to swallow, particularly needing `mem::transmute`, but this should reliably provide the same output on every platform. I had to check if WASM's endianness, but it seems like it's always little endian, as is x86(_64).

bitshifter (changes requested):
I would recommend adding default implementations in src/core/traits/vector.rs for `FloatVector2`, `FloatVector3` and `FloatVector4`.  The SSE2 trait implementation would only need to be implemented if fma is available. Shouldn't require any transmutes.

src/vec3.rs
```diff
@@ -417,7 +417,7 @@ impl_vecn_scalar_bit_op_traits!(UVec3, u32, XYZU32);
 impl_vecn_bit_op_traits!(UVec3, XYZU32);
 
 #[test]
-fn test_vec3_private() {
+fn test_vec3_mul_add() {
```
bitshifter, line 420: Now that this is public, is would be good to move these to `tests/vec*.rs`. Not mandatory for this PR though, only if you feel like it. I can tidy it up later otherwise.

bitshifter (approved):

src/vec3.rs
james7132: Please do. I'm not sure how to port this over with the macro heavy tests there.

bitshifter:
Thanks for all your work on this feature @james7132 !

PR #278, opened 2022-03-22
Fix lerp accuracy issues

### LINKED ISSUE #275: lerp()'s implementation has a loss of precision when inputs span multiple orders of magnitude
The general implementation of `lerp` in glam for floating point types takes the form as follows:

```rust
fn lerp(a: Self, b: Self, t: f32) -> Self {
  a + (b - a) * t
}
```

However, this presents floating point precision issues if `a` and `b` have very different exponent values due to the `(b - a)` in the computation. You can test this via `lerp(-16.0e30, 16.0, 1.0)` returning 0.0 instead of the correct 16.0. It may be more accurate to use the following form:

```rust
fn lerp(a: Self, b: Self, t: f32) -> Self {
  a * (1.0 - t) + b * t
}
```

This may not be a significant regression in performance as floating point multiplication is generally much faster than addition.

james7132:
Fixes #275 using the proposed methodology.

### DIFF
--- a/src/vec.rs
+++ b/src/vec.rs
@@ -416,7 +416,7 @@ macro_rules! impl_vecn_float_methods {
         #[doc(alias = "mix")]
         #[inline]
         pub fn lerp(self, other: Self, s: $t) -> Self {
-            self + ((other - self) * s)
+            self * (1.0 - s) + other * s
         }
 
         /// Returns true if the absolute difference of all elements between `self` and `other` is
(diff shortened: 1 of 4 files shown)


bitshifter (changes requested):
I left some comments on #275 - I would prefer to add an additional `lerp_precise` method than modify the existing `lerp` method. Primarily because the existing `lerp` method can be used to extrapolate and if people are using it that way this would break their code. A separate method means the precise option is there for those who need or want it.

Documentation would need to be added for the new and old methods explaining the difference.

It would also be good to add `glam_assert` checks to ensure `t` is in a valid range for the precise version.

bitshifter:
No changes here for a while, closing for now.

PR #281, opened 2022-04-03
Fix mistake in documentation for min() in vec.rs

AnEnigmaticBug:
I noticed that `min`'s documentation mentioned that it is equivalent to element-wise `max` which is incorrect. This small commit fixes that.


bitshifter:
Thanks!

Issue #276, opened 2022-03-05
[SSE2] Divide-by-zero in `Quat::slerp()` with TAU rotation


djeedai:
Minimal repro:

```rust
use glam::*;

fn main() {
    let q1 = Quat::IDENTITY;
    let q2 = Quat::from_rotation_x(std::f32::consts::TAU);
    let s = q1.slerp(q2, 1.);
    assert!(s.is_finite()); // Panics!
}
```

The issue is a divide by zero, because `q1` and `q2` represent the same rotation but are conjugate of each other, so `theta_sin` is all zeros here:

https://github.com/bitshifter/glam-rs/blob/0eb9b3bfc062c0f4a2f9223775c2abe0c8b89d25/src/core/sse2/quaternion.rs#L71

bitshifter:
Thanks for the report. This appears to be an issue for the SSE2 implementation, but not the scalar implementation.

bitshifter:
The slerp implementation was contributed by @aloucks in https://github.com/bitshifter/glam-rs/pull/57. According to the PR description it was based on the implementation in cgmath, but what is in glam differs from cgmath - https://github.com/rustgd/cgmath/blob/master/src/quaternion.rs#L135-L158.

aloucks:
I think cgmath was updated after #57.

https://github.com/rustgd/cgmath/blob/ff1cda766d48797d513f435b53c22c0a82718f5f/src/quaternion.rs#L129-L154

aloucks:
I think the issue is that `theta` evaluates to `PI` when the quaternion is created with a rotation of `TAU`. Both the current and old implementations fail (either `scale2` becomes zero or `theta_sin` becomes zero) when using the SSE `sin` approximation. When using the scalar `sin`, you'll get a very small non-zero value instead. 

@djeedai Does it work as expected when you force the dot product to be [positive](https://docs.rs/glam/0.20.2/glam/f32/struct.Quat.html#method.slerp)?

```rust
use glam::Quat;

fn main() {
    let q1 = Quat::IDENTITY;
    let q2 = Quat::from_rotation_x(std::f32::consts::TAU);
    let s = if q2.dot(q1) < 0.0 {
        q1.slerp(-q2, 1.)
    } else {
        q1.slerp(q2, 1.)
    };
    assert!(s.is_finite()); // No longer Panics!
}
```

I also have a branch that will check the dot product automatically: 

https://github.com/aloucks/glam-rs/tree/issue-276

PR #282, opened 2022-04-09
Ensure a positive dot product in `Quat::slerp`

aloucks:
Fixes https://github.com/bitshifter/glam-rs/issues/276.

This also has the benefit of no longer requiring the user to check
 the dot product and negating one of the quaternions if the dot
product is negative.


bitshifter:
Thank you very much for fixing that!

PR #288, opened 2022-05-14
Fix some unsound macros

SkiFire13:
These macros expanded the input expressions inside the `unsafe` block, which means user could do unsafe-only operations without writing `unsafe`. For example this compiles right now:

```rust
unsafe fn some_random_unsafe_fn() {}
glam::const_vec3!({
    some_random_unsafe_fn();
    [1.0; 3]
});
```

In some cases the macros themself were exploiting this unsoundess issue to do some union-transmuting when calling other macros, for example the one-argument version of `const_mat2 `. In those cases I had to add a new `unsafe` block.

This PR fixes this soundness issue by expanding the expression outside the `unsafe` block, assigning it to a local variable, and then using that one inside the `unsafe` block. The double braces are necessary because now the macros would expand to multiple statements/expressions, but they must expand to only one, so everything is wrapped in a block.

Also note that this slightly changes behaviour in some cases since previously the argument expression was evaluated more than one (see `const_vec3a!` for example), while now is it evaluated only once. I find this behaviour more intuitive than the old one, and probably nobody will notice the difference since you would need to trigger side-effects in that expression (like calling some function). Still, this could technically be considered a breaking change.


bitshifter:
Would you be able to add some tests that demonstrate the issue that this fixes?

SkiFire13:
I've added a `compile_fail` test with `trybuild`. Without the first commit it fails with:

```
test tests/macros_soundness/should_fail.rs [should fail to compile] ... error
Expected test case to fail to compile, but it succeeded.

test check_macros_soundness ... FAILED
```

I've also added a similar test but with a safe function as a proof that the errors all come from the `unsafe fn` call.

AFAIK there's also a way to do this without importing a third-party crate, but it requires using private doc tests which I find incredibly hacky.

bitshifter:
@SkiFire13 it seems that there was some slight difference in compiler output which caused it to fail, although I couldn't see what.

I would be OK with removing the tests again, probably not worth the trouble of introducing a new dep and potential brittleness.

bitshifter:
Thanks!

PR #294, opened 2022-06-13
Generate glam source

bitshifter:
Replace macro codegen with code generated offline by a tool. 

Improves readability of source code for users.


alice-i-cecile:
Spotted some [weird white space](https://github.com/bitshifter/glam-rs/pull/294/files#diff-5f704bf12495e4e932831ad7e138631214c2c6da7a1c27a332ed837a3daacb1cR1) in one of the generated files: perhaps run through a pass of `cargo fmt` as part of the generation?

bitshifter:
@alice-i-cecile for some reason GitHub didn't want to open that link. I do run rustfmt on the generated code, it doesn't remove empty lines though. White space can be controlled from the templates that are used to generate the source though, so it's just a case of fixing up the templates and regenerating. I'm doing a bunch more work separate to this PR though so I intend to try clean up spacing issues as part of that effort.

README.md
```diff
@@ -2,7 +2,7 @@
 
 [![Build Status]][github-ci] [![Coverage Status]][coveralls.io]
 [![Latest Version]][crates.io] [![docs]][docs.rs]
-[![Minimum Supported Rust Version]][Rust 1.52]
+[![Minimum Supported Rust Version]][Rust 1.57]
```
SkiFire13, line 5: `union` casts in `const fn`s were stabilized in Rust 1.56, so as a follow up to this PR many function could become `const fn`. cc #76

bitshifter: Oh, interesting. Thanks for the heads up.

src/f32/sse2/vec3a.rs
```diff
@@ -0,0 +1,1098 @@
...
+    ///
+    /// Using `mul_add` *may* be more performant than an unfused multiply-add if the target
+    /// architecture has a dedicated fma CPU instruction. However, this is not always true,
+    /// and will be heavily dependant on designing algorithms with specific target hardware in
+    /// mind.
+    #[inline]
+    pub fn mul_add(self, a: Self, b: Self) -> Self {
+        #[cfg(target_feature = "fma")]
+        unsafe {
+            _mm_fmadd_ps(self, b, c)
```
MarijnS95, line 616: Getting a compiler error in the `0.21` release here because `c` isn't defined :)

bitshifter:
I could do. Mostly kept the `.rs` extension so I'd get some kind of syntax highlighting. I could add a new extension to my editor though.

eddyb:
> I could do. Mostly kept the `.rs` extension so I'd get some kind of syntax highlighting.

Just saw this somewhat randomly, and wanted to add that [there's a neat trick](https://github.com/rust-lang/rust/blob/ca4e39400ef33198e2715973d1c67a1d3cee15e7/.gitattributes#L7-L8) to get at least GitHub to highlight certain extensions differently, e.g. you could add a `.gitattributes` containing:
```gitattributes
*.rs.tera linguist-language=Rust
```
(Official documentation for this GitHub feature [can be found in the `github/linguist` repo](https://github.com/github/linguist/blob/master/docs/overrides.md))

bitshifter:
Thanks for the tip @eddyb!

PR #298, opened 2022-06-22
Add `codegen` crate to `[workspace]` to allow running with `cargo r -p codegen`

MarijnS95:
This also makes the CI (i.e. `cargo fmt` and `cargo clippy`) pick up the crate.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
       # run clippy to verify we have no warnings
       - run: cargo fetch
       - name: cargo clippy
-        run: cargo clippy --all-features -- -D warnings
+        run: cargo clippy --workspace --all-features -- -D warnings
 
   test:
     name: Test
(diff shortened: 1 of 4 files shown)


bitshifter:
I had a feeling clippy would fail on codegen. I'll fix that up then merge this.

bitshifter:
I guess `test_no_std` should be excluded from the workspace? I'm not sure. This crate is to catch some weird no_std errors that only get caught when glam is a dependency, thus the separate crate. Usually you don't want to build it, just on CI really.

MarijnS95:
I think it's a good idea to have it _in_ the workspace. Tools like `rust-analyzer` will pick it up automatically (otherwise it may not get linted) and users running commands with `--all`/`--workspace` will automatically get the benefits of this test too.

MarijnS95:
Hope you don't mind some of the drive-by fixes made here :)

bitshifter:
All good stuff, thanks!

PR #303, opened 2022-06-22
ci: Run codegen and make sure changes are up-to-date

MarijnS95:


bitshifter:
Thanks. I was intending to add an option to codegen to do something like this, but this will do for now.

PR #307, opened 2022-06-24
macros: Add missing or fix broken `$crate::` prefix in (deprecated) `macro_rules! const_*`

MarijnS95:
This prefix was present in `glam 0.20` but is gone on `0.21.x`, now "requiring" the target type to be in scope - which is a reasonable assumption to make but not always true and a bit harsh to "force" the deprecation on uses (accidentally :)).

@bitshifter you probably want to backport this to 0.21 :)


bitshifter:
Ah thanks, that was an unintended change. I'll try get another release out with this in it today.

PR #308, opened 2022-06-24
codegen: Simplify `match Err + bail!` reporting with `.context()` and `?` operator

MarijnS95:
As mentioned in [#299] these verbose `Err` unwrapping structs accidentally slept in thanks to `anyhow` being pulled in with `default-features = false`; default features are now enabled which allow `anyhow` to do what it was designed for: turn any `Error` type into an `anyhow::Error` for painless propagation/bubbling of many different error types.  As a bonus we can use their `Context` trait to rewrap errors with more information where necessary, such as file paths or describing the operation at hand (as was already done manually in the `bail!` macros, while `Display`-formatting the underlying error).

[#299]: https://github.com/bitshifter/glam-rs/pull/299#discussion_r904008420

### DIFF
--- a/codegen/src/main.rs
+++ b/codegen/src/main.rs
@@ -1,10 +1,8 @@
 mod outputs;
 
-use anyhow::bail;
+use anyhow::{bail, Context};
 use clap::{arg, command};
 use rustfmt_wrapper::rustfmt;
-use std::fs::File;
-use std::io::Write;
 use std::path::Path;
 
 use outputs::build_output_pairs;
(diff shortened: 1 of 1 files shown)


codegen/src/main.rs
```diff
@@ -65,26 +47,22 @@ fn main() -> anyhow::Result<()> {
     let verbose = matches.is_present("verbose");
 
     if stdout && output_path_glob.is_none() {
+        // TODO: What if the glob matches multiple files?
```
bitshifter, line 50: This is a hangover from before I was using globs. I didn't really end up using this feature, it was easier to write output to a file to inspect it. Maybe it's useful to keep, but probably it would be better if I move this check until after the list of files has been generated and check that there's only one.

bitshifter:
Thanks, that's an awesome improvement :)

codegen/src/main.rs
MarijnS95: Oops, I forgot to ask you about this and remove the todo again before having it merged. Sounds like you have some plans to proceed and clean it up though :)

bitshifter:
This fixes a bug I was having where I didn't get any info if there was a Tera render error as well!

MarijnS95:
@bitshifter right, yes, `anyhow` recursively `Display`-prints the `source()`! (Whereas `{}` only `Display`s the current error, which implementation doesn't include `source()` to not break the former convention).

For those wondering what you just meant (I didn't process "any info" and thought there was no error at all), this is the before:

```
Error: tera error encountered: Failed to render 'mat.rs.tera'
```

And after:

```
Error: tera render error

Caused by:
    0: Failed to render 'mat.rs.tera'
    1: Variable `self_t` not found in context while rendering 'mat.rs.tera'
```

PR #309, opened 2022-06-28
Derive PartialEq and Eq when possible

Neo-Zhixing:
This PR derives the `PartialEq` and `Eq` trait when possible, instead of always implementing them. This is so that the integer vector types can be used as a constant generic - Rust requires const generic params to derive `Eq` and `PartialEq` in the unstable `adt_const_params` feature.

### DIFF
--- a/codegen/templates/vec.rs.tera
+++ b/codegen/templates/vec.rs.tera
@@ -135,7 +135,16 @@ pub const fn {{ self_t | lower }}(
 {%- else %}
 /// A {{ dim }}-dimensional vector.
 {%- endif %}
-#[derive(Clone, Copy)]
+#[derive(
+    Clone,
+    Copy,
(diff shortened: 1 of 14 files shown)


codegen/templates/vec.rs.tera
```diff
@@ -134,8 +134,19 @@ pub const fn {{ self_t | lower }}(
 /// This type uses 16 byte aligned SIMD vector type for storage.
 {%- else %}
 /// A {{ dim }}-dimensional vector.
-{%- endif %}
-#[derive(Clone, Copy)]
+{%- endif %}{% if not is_float %}
```
bitshifter, line 137: It doesn't matter for this PR, more FYI,  you can add a hyphen inside `{%-` to trim preceding white-space which would allow you to have this`if` block on a separate line.

PR #318, opened 2022-08-03
Simd quickfix

sammysheep:
- reverts your commit that removed support for std::simd from tests and readme
- updates std::simd API usage to the most recent:
  - simd_* instead of lanes_*
  - simd_min and simd_max instead of min for vertical min/max
  - splat is no longer const but there is a work around with `Simd::from_array` 

Passes:
```{bash}
$ cargo test --features core-simd
... elided ...
test result: ok. 204 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s

   Doc-tests glam

running 5 tests
test src/lib.rs - (line 115) ... ok
test src/lib.rs - (line 133) ... ok
test src/lib.rs - (line 64) ... ok
test src/lib.rs - (line 165) ... ok
test src/lib.rs - (line 202) ... ok

test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.20s
```

Please verify on your machine before merging.

### DIFF
--- a/src/f32/coresimd/vec3a.rs
+++ b/src/f32/coresimd/vec3a.rs
@@ -32,16 +32,16 @@ pub struct Vec3A(pub(crate) f32x4);
 
 impl Vec3A {
     /// All zeroes.
-    pub const ZERO: Self = Self::splat(0.0);
+    pub const ZERO: Self = Self(Simd::from_array([0.0; 4]));
 
     /// All ones.
-    pub const ONE: Self = Self::splat(1.0);
+    pub const ONE: Self = Self(Simd::from_array([1.0; 4]));
 
     /// All negative ones.
-    pub const NEG_ONE: Self = Self::splat(-1.0);
+    pub const NEG_ONE: Self = Self(Simd::from_array([-1.0; 4]));
 
     /// All NAN.
-    pub const NAN: Self = Self::splat(f32::NAN);
+    pub const NAN: Self = Self(Simd::from_array([f32::NAN; 4]));
 
     /// A unit-length vector pointing along the positive X axis.
     pub const X: Self = Self::new(1.0, 0.0, 0.0);
@@ -73,7 +73,7 @@ impl Vec3A {
     /// Creates a vector with all elements set to `v`.
     #[inline]
     pub const fn splat(v: f32) -> Self {
-        Self(f32x4::splat(v))
+        Self(Simd::from_array([v; 4]))
     }
 
     /// Creates a vector from the elements in `if_true` and `if_false`, selecting which to use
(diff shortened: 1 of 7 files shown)


src/f32/coresimd/vec3a.rs
```diff
@@ -32,16 +32,16 @@ pub struct Vec3A(pub(crate) f32x4);
 
 impl Vec3A {
     /// All zeroes.
-    pub const ZERO: Self = Self::splat(0.0);
+    pub const ZERO: Self = Self(Simd::from_array([0.0; 4]));
```
bitshifter, line 35: These should still be able to use `Self::splat` if `Self::splat` is still `const fn` (which `Vec3A::splat` is)

bitshifter (changes requested):
Most of the source files are generated. So for these fixes you need to change the template files in `codegen/templates/vec.rs.tera` and then run `cargo run -p codegen` to regenerate the sources. That should result in just the core-simd sources containing the right changes.

See https://github.com/bitshifter/glam-rs/blob/main/codegen/README.md.

bitshifter:
Thanks for the PR, I had a quick look at the core simd docs and couldn't find a replacement min. Do you know if there's a support channel somewhere that's a good place to ask questions about core simd?

src/f32/coresimd/vec3a.rs
sammysheep: I spoke to the maintainers and they had changed splat constness and this was the recommended workaround.

sammysheep:
> Thanks for the PR, I had a quick look at the core simd docs and couldn't find a replacement min. Do you know if there's a support channel somewhere that's a good place to ask questions about core simd?

It's now called `simd_min`. It is changed in the code I submitted methinks. See:
https://rust-lang.github.io/portable-simd/core_simd/trait.SimdFloat.html#tymethod.simd_min

Same for simd_max.

sammysheep:
I wouldn't call it support, but the maintainers are very nice and you can chat with them on Zulip:
https://rust-lang.zulipchat.com/#narrow/stream/257879-project-portable-simd

sammysheep:
> Most of the source files are generated. So for these fixes you need to change the template files in `codegen/templates/vec.rs.tera` and then run `cargo run -p codegen` to regenerate the sources. That should result in just the core-simd sources containing the right changes.
> 
> See https://github.com/bitshifter/glam-rs/blob/main/codegen/README.md.

Sorry, I missed that. Should I give it a shot in this PR, or do you want to make the changes?

sammysheep:
Okay, I altered the template and removed the explicit splatting for the vars so it instead uses the struct `impl` splat. The codegen'd changes are testing correctly.

 I think maybe that is closer to what you wanted if you want to run your workflow again.

bitshifter (approved):

bitshifter:
Thanks for that!

PR #322, opened 2022-08-14
Add Borsh serializer support

cybersoulK:


bitshifter:
Hi there, 

I prefer it if people create a discussion to ask if new features are wanted to avoid wasting anyone's time, see https://github.com/bitshifter/glam-rs/blob/main/CONTRIBUTING.md#new-functions-and-methods. I should really make a pull request template because no one reads CONTRIBUTING in my experience. Also with a PR it's good to leave a comment saying what the PR is for and why it would benefit glam users.

In this case of this crate, I don't want to add support for it for a few reasons:

* Primarily because it is part of the web3 ecosystem (https://near.org/), while the crate itself doesn't appear to have anything to do with web3 or blockchain the association is something I want to avoid as maintainer of `glam`
* `glam` already provides serialisation support through `serde` and `rykv` which are widely used in the Rust game dev ecosystem
* The number of users wanting to use this crate is small (e.g. one person AFAIK verses `serde` and `rykv` where they are used by popular Rust game engines)
* There are a lot of serialization crates, I don't want to support them all, even if maintenance cost is reasonably low.

Hopefully at some stage the orphan rule will be relaxed and it will be easier to implement traits without needing newtypes, at which point there will be less burden on crate maintainers to add support for many many different crates.

cybersoulK:
borsh it he only one having enums be 1 byte, i tried 5 different crates.
that increases throughput by 25% in multiplayer games. 
just letting that out there

cybersoulK:
my code that i offered is fairly straight forward, and it just works. no need to do anything else

luxalpa:
Maybe forks like these could be referenced in the Readme file to make them easier to find? I'm currently adding `speedy` support to my version of this crate, using @cybersoulK's version of the code as a help for my own implementation.

PR #333, opened 2022-08-31
Implement Sum<Self>, Product<Self> for *VecN, *MatN, Quat

nekodjin:


bitshifter (commented):
It would be good to update the current tests as well, since I'm not clear on what code the reference implementations don't work for.

bitshifter:
I'd prefer it if tests are included with the PR, or at least make it clear how this can be tested.

For example the current Vec2 sum test effectively (with macros removed) does:

```rust
assert_eq!(vec![Vec2::ONE, Vec2::ONE].iter().sum::<Vec2>(), Vec2::ONE + Vec2::ONE);
```

which seems fine with the sum impl that takes a ref, it would be good to see a example like the above that doesn't work with the existing ref sum impl, I haven't seen a complete code example yet which exhibits the problem.

bitshifter:
Answering my own question, this should hopefully be sufficient to test the new impl:

```rust
assert_eq!(vec![Vec2::ONE, Vec2::ONE].into_iter().sum::<Vec2>(), Vec2::ONE + Vec2::ONE);
```

Issue #339, opened 2022-09-26
Add const to from_diagonal


VirxEC:
The function -> https://docs.rs/glam/0.21.3/glam/f32/struct.Mat3A.html#method.from_diagonal
Looking at the code, there's no real reason why this function can't be a `const fn`. Can this please be changed?

Also, it looks like `to_cols_array_2d` can also easily be a `const fn`.

VirxEC:
I'll make a pr then

bitshifter:
Thanks for the PR!

PR #340, opened 2022-09-27
Make from_diagonal a const fn

VirxEC:
First PR! I was unsure if I should update all references outside of `src/` so I didn't.

The one oddity is `src/f32/sse2/mat4.rs`'s `from_diagonal`. I had to get the x/y/z/w via `let [x, y, z, w] = diagonal.to_array();` because I found `diagonal.x`, `diagonal.y`, etc can't be done in a const contexts for SIMD-enabled vectors like `Vec3A` and `Vec4`.

I ran `build_and_test_features.sh` and there were no errors over the various tests.

### DIFF
--- a/codegen/templates/mat.rs.tera
+++ b/codegen/templates/mat.rs.tera
@@ -334,11 +334,30 @@ impl {{ self_t }} {
             {%- endfor %}
         ]
     }
-
+{% if dim == 4 and vecn_t != "DVec4" %}
     /// Creates a {{ nxn }} matrix with its diagonal set to `diagonal` and all other entries set to 0.
     #[doc(alias = "scale")]
     #[inline]
-    pub fn from_diagonal(diagonal: {{ vecn_t }}) -> Self {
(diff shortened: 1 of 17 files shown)


VirxEC:
Closes #339 when merged

bitshifter:
The rustfmt error is weird since rustfmt is run on the codegen output. You could add a `#[rustfmt::skip]` before that block and ignore it. The wasm and core-simd features aren't compiled by the main build script, there are separate scripts for building and testing wasm. Possibly I haven't added one for core-simd but you can just specify it as a feature (requires nightly).

VirxEC:
I think I fixed it

VirxEC:
I changed the codegen, so fingers crossed, it should pass all tests now.

bitshifter:
What version of rust are you running codegen with btw? I guess a different version of rustfmt might generate different results. Usually I use the latest stable.

VirxEC:
The thing is that I wasn't, until the last commit 😜

VirxEC:
Yep! All the tests passed now :D

PR #342, opened 2022-09-29
Add BVec::splat

rodolphito:
For API parity with other Vecs, add tests, and opportunistically fix a doc typo. I don't absolutely *need* this function, but it would make my code nicer.

### DIFF
--- a/codegen/templates/vec_mask.rs.tera
+++ b/codegen/templates/vec_mask.rs.tera
@@ -125,7 +125,49 @@ impl {{ self_t }} {
         {% endif %}
     }
 
-    /// Returns a bitmask with the lowest two bits set from the elements of `self`.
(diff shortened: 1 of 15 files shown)


codegen/templates/vec_mask.rs.tera
```diff
@@ -125,7 +125,49 @@ impl {{ self_t }} {
         {% endif %}
     }
 
-    /// Returns a bitmask with the lowest two bits set from the elements of `self`.
+    /// Creates a vector with all elements set to `v`.
+    #[inline]
+    pub const fn splat(v: bool) -> Self {
+        {% if is_scalar and is_bool %}
```
bitshifter, line 131: This could call `Self::new` just to reuse existing code, it's `#[inline(always)]` so there shouldn't be any overhead.

PR #343, opened 2022-09-29
Add more const fn's

VirxEC:
Because I'm bored-ish and this is kinda fun.

FYI I'm now testing with `cargo run --release -p codegen -- -f && ./build_and_test_features.sh && ./build_and_test_wasm32_chrome.sh` so I'm not testing coresimd (famous last words)

Wondering if this kind of thing is wanted or if the extra complexity isn't welcome :thinking:

### DIFF
--- a/codegen/templates/vec.rs.tera
+++ b/codegen/templates/vec.rs.tera
@@ -372,7 +372,7 @@ impl {{ self_t }} {
 {% if dim == 2 %}
     /// Creates a 3D vector from `self` and the given `z` value.
     #[inline]
-    pub fn extend(self, z: {{ scalar_t }}) -> {{ vec3_t }} {
+    pub const fn extend(self, z: {{ scalar_t }}) -> {{ vec3_t }} {
         {{ vec3_t }}::new(self.x, self.y, z)
     }
 {% elif dim == 3 %}
--- a/codegen/templates/vec.rs.tera
+++ b/codegen/templates/vec.rs.tera
@@ -389,8 +389,9 @@ impl {{ self_t }} {
 
     /// Creates a 4D vector from `self` and the given `w` value.
     #[inline]
-    pub fn extend(self, w: {{ scalar_t }}) -> {{ vec4_t }} {
-        {{ vec4_t }}::new(self.x, self.y, self.z, w)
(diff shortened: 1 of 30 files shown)


codegen/templates/vec.rs.tera
```diff
@@ -389,8 +389,9 @@ impl {{ self_t }} {
 
     /// Creates a 4D vector from `self` and the given `w` value.
     #[inline]
-    pub fn extend(self, w: {{ scalar_t }}) -> {{ vec4_t }} {
-        {{ vec4_t }}::new(self.x, self.y, self.z, w)
+    pub const fn extend(self, w: {{ scalar_t }}) -> {{ vec4_t }} {
```
bitshifter, line 392: Probably I should provide a SIMD implementation of Vec3A extend, which might be faster. It would mean this couldn't be const.

src/f32/coresimd/mat2.rs
```diff
@@ -71,8 +76,9 @@ impl Mat2 {
     /// Creates a `[[f32; 2]; 2]` 2D array storing data in column major order.
     /// If you require data in row major order `transpose` the matrix first.
     #[inline]
-    pub fn to_cols_array_2d(&self) -> [[f32; 2]; 2] {
-        [self.x_axis.to_array(), self.y_axis.to_array()]
+    pub const fn to_cols_array_2d(&self) -> [[f32; 2]; 2] {
+        let self_ = unsafe { &*(self as *const Self as *const crate::deref::Cols2<Vec2>) };
```
bitshifter, line 80: I think this could cast directly to `[[f32; 2]; 2]`.

src/f32/coresimd/mat2.rs
```diff
@@ -56,8 +61,8 @@ impl Mat2 {
     /// Creates a `[f32; 4]` array storing data in column major order.
     /// If you require data in row major order `transpose` the matrix first.
     #[inline]
-    pub fn to_cols_array(&self) -> [f32; 4] {
-        [self.x_axis.x, self.x_axis.y, self.y_axis.x, self.y_axis.y]
+    pub const fn to_cols_array(&self) -> [f32; 4] {
+        unsafe { UnionCast { v: *self }.a }
```
bitshifter, line 65: This could probably also cast directly to `[f32; 4]`

PR #344, opened 2022-09-29
Fix negation and signum to consistently handle negative zero across platforms

rodolphito:
The implementation gave different results on different target platforms when working with negative zero.


bitshifter:
Just needs a rustfmt I think

PR #345, opened 2022-09-29
Add Vec::sign_bits

rodolphito:
This is dependent on the fixes in #344 (those problems were found while adding this function)

I need this function for optimized tetrahedron face winding order checks (four dot products being put in a `Vec` and then grabbing the sign bits to check if its either `0b0000` or `0b1111`, thus having all the same winding order and not being degenerate) among other things.

`BVec::bitmask` almost does this, but gets the LSB for scalar implementations, but I need it on `Vec` anyways so its irrelevant

### DIFF
--- a/codegen/templates/vec.rs.tera
+++ b/codegen/templates/vec.rs.tera
@@ -831,7 +831,7 @@ impl {{ self_t }} {
     /// A negative element results in a `1` bit and a positive element in a `0` bit.  Element `x` goes
     /// into the first lowest bit, element `y` into the second, etc.
     #[inline]
-    pub fn sign_bits(self) -> u32 {
+    pub fn is_negative_bitmask(self) -> u32 {
         {% if is_scalar and is_float %}
             {% for c in components %}
                 {% if loop.first %}
--- a/codegen/templates/vec.rs.tera
+++ b/codegen/templates/vec.rs.tera
@@ -862,9 +862,9 @@ impl {{ self_t }} {
             {% endif %}
         {% elif is_coresimd %}
             {% if dim == 3 %}
-                (self.0.to_bitmask() & 0x7) as u32
+                (self.0.is_sign_negative().to_bitmask() & 0x7) as u32
             {% elif dim == 4 %}
-                self.0.to_bitmask() as u32
+                self.0.is_sign_negative().to_bitmask() as u32
             {% endif %}
         {% endif %}
     }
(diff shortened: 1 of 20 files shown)


bitshifter:
I'm wondering what the most general purpose method would be for most users here.

For example, there are already methods on `VecN` for `is_finiite -> bool` (all elements finite) and `is_nan -> bool` (any elements are nan). I've tried to match the naming use on `f32`/`f64` there although I guess it's not entirely consistent about whether the result is `any` or `all`. There is also `is_nan_mask` which returns `BVecN`.

For your usage it seems like a `is_sign_negative` which returns `true` if `all` elements are sign negative would fit your use case? That would probably fit better with the existing naming scheme. Do you need a `u32` mask beyond checking if it's all zeros or all ones?

rodolphito:
My use-case needs to distinguish between three cases: all positive, all negative, or mixed signs. I don't need to know what the specific mix is if its mixed. Sample code:

```
    match sign_winding.sign_bits();
    {
        // All positive
        0 => sign_origin.cmpge(Vec4::splat(-f32::EPSILON)),

        // All negative
        0xf => sign_origin.cmple(Vec4::splat(f32::EPSILON)),

        // Mixed signs, degenerate tetrahedron
        _ => BVec4A::splat(true)
    }
```

codegen/templates/vec.rs.tera
```diff
@@ -825,6 +825,49 @@ impl {{ self_t }} {
             }
         {% endif %}
     }
+
+    /// Returns a bitmask with the lowest {{ dim }} bits set to the sign bits from the elements of `self`.
+    ///
+    /// A negative element results in a `1` bit and a positive element in a `0` bit.  Element `x` goes
+    /// into the first lowest bit, element `y` into the second, etc.
+    #[inline]
+    pub fn sign_bits(self) -> u32 {
```
bitshifter, line 834: I think perhaps `is_negative_bitmask` would be a better name, `sign` can be positive or negative and I already use `mask` in a few places to refer to a vector mask (e.g. `BVecN`).

bitshifter:
Kind of a note to myself, it might be good to add `TRUE` and `FALSE` constants to `BVecN` types for this kind of use.

rodolphito:
Funny, there's a FALSE const for all BVecN types but no TRUE. I can PR this if its convenient

bitshifter:
The current false is private I think. If you want to make a PR, go for it.

PR #363, opened 2022-11-27
Fix Changelog

dallenng:
I read the changelog and I was surprised to see the addition of `look_at_lh` and `look_at_rh` because I was already using those functions a long time ago. I dug the commits a bit and saw the addition of `look_to*` functions rather than `look_at*`.


bitshifter:
Thanks! Good to know people read the change log :)

PR #368, opened 2022-12-18
Make BVec4A available when scalar-math feature is enabled

rodolphito:
This just matches what was done for BVec3A. There's a bunch of functions in scalar/vec4.rs that return BVec4 instead of BVec4A though, like `cmpge`. I tried wrangling the template files but wasn't sure how to fix that part, its a bit annoying because switching on scalar-math creates compile errors when these functions are being used. I currently only need scalar-math to run tests in miri because miri doesn't support SIMD.

### DIFF
--- a/src/bool.rs
+++ b/src/bool.rs
@@ -68,12 +68,14 @@ pub use coresimd::bvec4a::BVec4A;
 ))]
 pub use scalar::bvec3a::BVec3A;
 
-#[cfg(not(any(
-    feature = "scalar-math",
-    feature = "core-simd",
-    target_feature = "sse2",
-    target_feature = "simd128"
-),))]
+#[cfg(any(
+    not(any(
+        feature = "core-simd",
+        target_feature = "sse2",
+        target_feature = "simd128"
+    )),
+    feature = "scalar-math"
+))]
 pub use scalar::bvec4a::BVec4A;
 
 mod const_test_bvec2 {


bitshifter:
The clippy lint error looks unrelated, could just be something that's been turned on in a new Rust version.

bitshifter:
What compile errors are you getting with scalar-math enabled? `Vec3A` is a bit different to `Vec4`. It doesn't really make sense to return `Vec4A` with `scalar-math` enabled and it doesn't really make sense to use `Vec3A` at all with scalar math enabled.

rodolphito:
`Vec4::ONE.cmpge(Vec4::ZERO)` returns a `BVec4A` usually, but with `scalar-math` enabled it returns a `BVec4`. Same with all the other comparison functions. I think using `BVec4A` makes a lot of sense since its just a bitmask and can fit well inside a usize, even a single byte is enough. I'm not fully sure there's even a difference between `BVec4A` and `BVec4`.

Example conditional compilation error when enabling `scalar-math`:

```rs
fn origin_outside_of_tetrahedron_planes(a: Vec3A, b: Vec3A, c: Vec3A, d: Vec3A) -> BVec4A {
    // snipped
    let sign_bits: u32 = signd.is_negative_bitmask();
    match sign_bits {
        // All positive
        0 => signp.cmpge(Vec4::splat(-f32::EPSILON)),

        // All negative
        0xf => signp.cmple(Vec4::splat(f32::EPSILON)),

        // Mixed signs, degenerate tetrahedron
        _ => BVec4A::TRUE,
    }
}
```

bitshifter:
`BVec4A` definitely performs worse when SIMD is not available, so for `scalar-math` users I would prefer to stick with `BVec4`.

You can however run miri by disabling the sse2 target feature. which hopefully will solve the problem you are having.

Example command line:

```sh
RUSTFLAGS='-C target-feature=-sse2' cargo +nightly miri test
```

Issue #122, opened 2021-01-20
Bytemuck traits for SIMD types


agorgl:
Currently plain (e.g. Vec3) types have Pod and Zeroable traits from bytemuck implemented but simd types (e.g. Vec3A) have not.
Implementing it in them would allow us to use them easily in Uniform and Buffer structs to be used directly with the GPU.

bitshifter:
My read is that`Vec3A` is not eligible to be `Pod` in bytemuck due to padding at the end (because of alignment). Technically it would be eligible if simd is enabled but not otherwise. See https://docs.rs/bytemuck/1.5.0/bytemuck/trait.Pod.html

sotrh:
Looking at the docs, it seems that using `align` with `repr(C)` is one of the exceptions to the no padding rule.

> The type needs to be repr(C) or repr(transparent). In the case of repr(C), the packed and align repr modifiers can be used as long as all other rules end up being followed.

From the definition of `XYZF32A16` it seems like it fits the bill.

```rust
/// Ths XYZF32A16 is used for the Vec3A type, that is a 16 btye aligned XYZ<f32> type.
#[derive(Clone, Copy, Debug, Default, PartialEq, PartialOrd)]
#[cfg_attr(target_arch = "spirv", repr(simd))]
#[cfg_attr(not(target_arch = "spirv"), repr(align(16), C))] // <-- here
pub struct XYZF32A16 {
    pub x: f32,
    pub y: f32,
    pub z: f32,
}
```

bitshifter:
The key part being 

> as long as all other rules end up being followed

which includes 

> The type must not contain any padding bytes, either in the middle or on the end 

So I don't think that is an exception at all. It is possible to have `packed` or `align` data that contains no internal padding and that is OK to be `bytemuck::Pod`, but `Vec3A` definitely contains padding.

emilk:
Why doesn't `Vec3A` contain an explicit padding?

i.e. change

```rust
pub struct Vec3A {
    pub x: f32,
    pub y: f32,
    pub z: f32,
}
```

to

```rust 
pub struct Vec3A {
    pub x: f32,
    pub y: f32,
    pub z: f32,
    pub _padding: f32,
}
```

I think that would solve this issue

bitshifter:
I did start playing around with this idea, I can't remember where I got to with it. I think I wanted to put it off until there's an aarch64 simd implementation (which is WIP) otherwise it would pessimise `Vec3A` performance a bit.

[...]

PR #370, opened 2022-12-25
add bytemuck support for Vec3A on simd128 and ssd2 feature

Alchemist0823:
add Vec3A bytemuck support on vectorized platform.


bitshifter:
This will result in an incompatible API between x86 and other platforms. The way this should be resolved is this https://github.com/bitshifter/glam-rs/issues/122#issuecomment-1282513894.

PR #374, opened 2023-02-03
Fix incorrect comment on `signum` function for integer vectors.

PennyMew:
The existing doc comment in `ivec2/3/4.rs` incorrectly states that the result of `IVec2/3/4::signum` will never have an element equal to 0. Additionally, it refers to floating point `+0.0`, `-0.0`, `INFINITY`, and `NEG_INFINITY`, which are not `i32` values. `IVec2`, `IVec3`, and `IVec4` all implement `signum` by calling the Rust standard `i32::signum` on each element of the vector. `i32::signum` returns 0 when the i32 value is 0. This PR modifies `vec.rs.tera` to generate correct comments for the integer vectors.

### DIFF
--- a/codegen/templates/vec.rs.tera
+++ b/codegen/templates/vec.rs.tera
@@ -801,9 +801,15 @@ impl {{ self_t }} {
 
     /// Returns a vector with elements representing the sign of `self`.
     ///
+    {% if is_float -%}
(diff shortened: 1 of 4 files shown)


bitshifter:
Thanks!

PR #384, opened 2023-03-13
depend on `serde_derive` directly

jakobhellermann:
If no crate enables the `derive` feature of `serde,` then serde and serde_derive can compile in parallel, and anything depending only on serde can compile while serde_derive is not yet done.
This can save up to 5 seconds of compilation, depending on whether this helps the critical chain and if no other crate enables the feature.

This cuts the critical chain of bevy compilation from `serde_derive -> serde -> hashbrown -> indexmap -> petgraph -> bevy_utils -> bevy_reflect` to `serde_derive -> glam -> bevy_math -> bevy_reflect`  which is faster.

image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -44,13 +44,16 @@ fast-math = []
 # experimental nightly portable-simd support
 core-simd = []
 
+serde = ["dep:serde", "serde_derive"]
+
 [dependencies]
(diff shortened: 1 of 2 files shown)


bitshifter:
Is `glam` the only thing in `bevy` that depends on `serde_derive`?

bitshifter:
Don't need this one now.

PR #385, opened 2023-03-13
remove `serde_derive`

jakobhellermann:
alternative to https://github.com/bitshifter/glam-rs/pull/384

Cuts the critical compilation chain of bevy further from `syn -> serde_derive -> glam -> bevy_math -> bevy_reflect` to `syn -> bevy_reflect_derive -> bevy_reflect` which is a few seconds faster.

**but** it makes the code less maintainable (especially if you plan to add more types with a serde derives so I can understand if you don't want to merge this. I just copied the output of `rust-analyzer`'s `Expand macro recursively` and cleaned it up a bit.


bitshifter:
Seems like this doesn't work with no_std.

bitshifter:
Since this was the only type using derive I'm OK with providing an implementation by hand.

It should move to `src/features/impl_serde.rs` with the rest of them though.

bitshifter (changes requested):
Looks good but this should move to `src/features/impl_serde.rs`.

bitshifter (approved):

PR #387, opened 2023-03-16
Calculate Quat::to_axis_angle in a more numerically stable way

jnises:
The current implementation results in unnormalized axis when the angle is close to 0.

This way it is done in this PR is probably a bit slower.
But handles angles close to 0 better.
It seems to be how Eigen and nalgebra does it for example.
It guarantees that the axis is normalized.

### DIFF
--- a/codegen/templates/quat.rs.tera
+++ b/codegen/templates/quat.rs.tera
@@ -456,18 +456,18 @@ impl {{ self_t }} {
         }
     }
 
-    /// Returns the rotation axis and angle (in radians) of `self`.
+    /// Returns the rotation axis (normalized) and angle (in radians) of `self`.
     #[inline]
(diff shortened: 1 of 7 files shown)


bitshifter:
Not sure why the codegen lint failed since it looks like you have run codegen. Possibly you need to run it again or maybe the version of Rust you are on is producing different codegen.

Otherwise this looks fine in principle.

jnises:
> Not sure why the codegen lint failed since it looks like you have run codegen. Possibly you need to run it again or maybe the version of Rust you are on is producing different codegen.
> 
> Otherwise this looks fine in principle.

I hade completely missed the template bit  :)
I was thinking about how error prone it was to update all those files manually 😅 
I still get codegen mismatches when I run that locally, but that is due to line endings it seems. Hopefully they match on CI.

bitshifter (approved):

bitshifter:
Thanks!

PR #389, opened 2023-04-02
Make enabling libm override std math functions

bitshifter:


Cargo.toml
```diff
@@ -29,7 +29,7 @@ glam-assert = []
 scalar-math = []
 
 # libm is required when building no_std
-libm = ["num-traits/libm"]
+#libm = ["num-traits/libm"]
```
bitshifter, line 32: remove

Issue #195, opened 2021-06-15
Audit `From` conversion implementations


bitshifter:
I am thinking of removing some `From` conversions, mostly the ones that perform a `truncate`. The main reason being that if these are used in generic code they might be surprising. 

One concrete example is I was looking at ways to make using `Vec3A` a bit nicer. I lot of methods take `Vec3` but if they were made to take `Into<Vec3>` then the user would not need to manually `.into()` the `Vec3A`. However this would also accept any of the `From` implementations for `Vec3`. So for this kind of use case it would be good to remove anything that might be "surprising".

bitshifter:
Current impls

```
src/affine2.rs|321 col 14| impl From<$affine2> for $mat3 {
src/affine2.rs|434 col 6| impl From<Affine2> for Mat3A {
src/affine3.rs|460 col 14| impl From<$affine3> for $mat4 {
src/mat2.rs|286 col 14| impl From<$mat3> for $mat2 {
src/mat3.rs|437 col 14| impl From<$mat4> for $mat3 {
src/mat3.rs|532 col 6| impl From<Mat3> for Mat3A {
src/mat3.rs|539 col 6| impl From<Mat3A> for Mat3 {
src/quat.rs|621 col 14| impl From<$quat> for $vec4 {
src/quat.rs|628 col 14| impl From<$quat> for ($t, $t, $t, $t) {
src/quat.rs|635 col 14| impl From<$quat> for [$t; 4] {
src/quat.rs|642 col 14| impl From<$quat> for $inner {
src/vec.rs|508 col 14| impl From<$vecn> for $inner {
src/vec.rs|515 col 14| impl From<$inner> for $vecn {
src/vec.rs|690 col 14| impl From<[$t; $size]> for $vecn {
src/vec.rs|697 col 14| impl From<$vecn> for [$t; $size] {
src/vec2.rs|140 col 14| impl From<($t, $t)> for $vec2 {
src/vec2.rs|147 col 14| impl From<$vec2> for ($t, $t) {
src/vec3.rs|143 col 14| impl From<($vec2, $t)> for $vec3 {
src/vec3.rs|150 col 14| impl From<($t, $t, $t)> for $vec3 {
src/vec3.rs|157 col 14| impl From<$vec3> for ($t, $t, $t) {
src/vec3.rs|164 col 14| impl From<$vec3> for $vec2 {
src/vec3.rs|318 col 6| impl From<Vec3> for Vec3A {
[...]

PR #392, opened 2023-04-12
Remove legacy `::from()` truncation mention in docs; add more intradoc links

MarijnS95:
`::from()` to truncate `VecN` to `Vec(N - 1)` has been removed long ago in #195, and should not be mentioned anymore in the docs.  Also replace more type and function references with proper intradocs, making it easier for the reader to click through to the thing that is being mentioned as well as validate such documentation (in particular: validity of intradoc links) on the CI in the future.

### DIFF
--- a/codegen/templates/vec.rs.tera
+++ b/codegen/templates/vec.rs.tera
@@ -147,10 +147,10 @@ pub const fn {{ self_t | lower }}(
 {%- if self_t == "Vec3A" %}
 ///
 /// SIMD vector types are used for storage on supported platforms for better
-/// performance than the `Vec3` type.
+/// performance than the [`Vec3`] type.
 ///
-/// It is possible to convert between `Vec3` and `Vec3A` types using `From`
-/// trait implementations.
+/// It is possible to convert between [`Vec3`] and [`Vec3A`] types using [`From`]
+/// or [`Into`] trait implementations.
 ///
 /// This type is 16 byte aligned.
 {%- elif self_t == "Vec4" and is_simd %}
@@ -412,7 +412,7 @@ impl {{ self_t }} {
 
     /// Creates a 2D vector from the `x` and `y` elements of `self`, discarding `z`.
     ///
-    /// Truncation may also be performed by using `self.xy()` or `{{ vec2_t }}::from()`.
+    /// Truncation may also be performed by using [`self.xy()`][crate::swizzles::Vec3Swizzles::xy()].
     #[inline]
     pub fn truncate(self) -> {{ vec2_t }} {
         use crate::swizzles::Vec3Swizzles;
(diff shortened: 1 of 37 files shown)


codegen/templates/vec.rs.tera
```diff
@@ -421,10 +421,10 @@ impl {{ self_t }} {
 {% elif dim == 4 %}
     /// Creates a 2D vector from the `x`, `y` and `z` elements of `self`, discarding `w`.
     ///
-    /// Truncation to `{{ vec3_t }}` may also be performed by using `self.xyz()` or `{{ vec3_t }}::from()`.
+    /// Truncation to [`{{ vec3_t }}`] may also be performed by using [`self.xyz()`][crate::swizzles::Vec4Swizzles::xyz()].
 {%- if scalar_t == "f32" %}
     ///
-    /// To truncate to `Vec3A` use `Vec3A::from()`.
+    /// To truncate to [`Vec3A`] use [`Vec3A::from()`].
```
MarijnS95, line 427: This isn't linking to the right type, we should probably have a `<Vec3A as From<{{ vec4_t }}>>::from()` for that but it is super verbose and doesn't work.

bitshifter: Perhaps a limitation in rustdoc?

bitshifter (approved):

bitshifter:
Thanks!

codegen/templates/vec.rs.tera
MarijnS95: Wouldn't be surprised, `<X as Y>::...` is rather complex to parse, and I don't think there's another easy way to point out exactly which trait `impl` _and trait `impl` `fn`_ is intended. Clicking the links on the latest docs.rs show that the `impl` is a relatable unique value, but the `from` just has a numeric suffix:

`impl From<Vec3> for Vec3A`:
https://docs.rs/glam/latest/glam/f32/struct.Vec3A.html#impl-From%3CVec3%3E-for-Vec3A
https://docs.rs/glam/latest/glam/f32/struct.Vec3A.html#method.from-8


`impl From<Vec3A> for Vec3`:
https://docs.rs/glam/latest/glam/f32/struct.Vec3A.html#impl-From%3CVec3A%3E-for-Vec3
https://docs.rs/glam/latest/glam/f32/struct.Vec3A.html#method.from

Don't think it's worth spending much time figuring this out, fully-qualified syntax seems to be the only way to solve this and that's not yet supported:
https://github.com/rust-lang/rust/issues/74563
https://rust-lang.github.io/rfcs/1946-intra-rustdoc-links.html#linking-to-associated-items
https://rust-lang.github.io/rfcs/1946-intra-rustdoc-links.html#traits-in-scope

PR #396, opened 2023-04-24
Add repr(C) to unions used for type punning

mattbork:
In the SSE2 code and a few other places, unions are used to initialize SIMD types from scalar arrays through type punning. However, the default representation for unions makes no guarantees about the size of the union or the offsets of its fields. Only with `#[repr(C)]` are all fields guaranteed to have an offset of 0 and thus fully overlap. It is hard to imagine how the compiler would ever deviate from this under `#[repr(Rust)]`, but as far as I understand them, these are the rules.


bitshifter:
Thanks!

PR #398, opened 2023-05-04
Use right amount of column vectors in documentation

stakka:
Just a simple documentation fix to correct an error most likely coming from a c/p mistake.


bitshifter (changes requested):
These files are all generated, the docs will need to be fixed in the templates used to generate them. See https://github.com/bitshifter/glam-rs/blob/main/codegen/README.md for details.

stakka:
I see, apologies for rushing to it. I hope this solution helps.

bitshifter (approved):
Looks good 👍

PR #409, opened 2023-06-14
doc: range for Vec2 angle_between

tatref:
I'm not sure if there's a preference between `π`, `&pi;`, `pi`...

### DIFF
--- a/codegen/templates/vec.rs.tera
+++ b/codegen/templates/vec.rs.tera
@@ -1447,7 +1447,7 @@ impl {{ self_t }} {
         }
     }
 
-    /// Returns the angle (in radians) between `self` and `rhs`.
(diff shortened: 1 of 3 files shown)


bitshifter:
Thanks for the PR. Most code in glam is generated, so for this change you need to update the vec template in codegen/templates then run the codegen tool. Otherwise looks ok, maybe "in the range" or "within the range" over "in range"?

tatref:
Thanks for the tip, I didn't know the code was generated, that's cool!

And sorry for my bad english :-)

bitshifter:
No worries, one last thing, the opening bracket of the range should be `[`, at the moment it's `]`.

tatref:
At first I thought it was not included.
Fixed now.

PR #411, opened 2023-06-22
Use libm::fabs() & libm::fabsf()

Patryk27:
libm's implementation is shorter + the current approach sometimes yields _very funky_ results on Windows (I haven't investigated it deeper though, since swapping the implementation to libm solves my problem).


bitshifter:
Thanks, for some reason I didn't think libm had a fabs implementation, not sure why.

PR #415, opened 2023-07-09
Updated int vec2 to have l1 distance and chessboard distance

Veliladon:
Hi,

I added L1 Distance and Chebyshev distance computation functions to all integer vecs as I've found it really useful in dealing with grid positions using integer Vec2s. They return the next biggest int size as a result because it's possible to have 2*MAX_INT as the result.

I tried to keep to the style and I added docs for it. Please let me know if you need me to do anything else and if these functions are not wanted please feel free to close this.

### DIFF
--- a/src/i32/ivec2.rs
+++ b/src/i32/ivec2.rs
@@ -286,6 +286,22 @@ impl IVec2 {
         (self - rhs).length_squared()
     }
 
+    /// Compute the L1 distance between two points in space.
+    #[inline]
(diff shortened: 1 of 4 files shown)


Veliladon:
I read the contributing docs properly and since it's all generated I'm closing this pull request as it's a not good way to do this.

bitshifter:
If you want to create a PR that uses the code generator I'm happy to consider it.

PR #416, opened 2023-07-14
Set `default-features = false` on rkyv

EmilioLaiso:
Using default features pulls in the `size_32` feature, but the choice of the `size_x` feature should be left to the top-level crate as it was thought of in https://github.com/bitshifter/glam-rs/pull/223#discussion_r683917210 and https://github.com/bitshifter/glam-rs/pull/223#issuecomment-895328926, and should be working now that `glam-rs` is on edition 2021 and rust >= 1.50.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Build-test documentation
         env:
           RUSTDOCFLAGS: -Dwarnings
-        run: cargo doc --all --all-features --no-deps --document-private-items
+        run: cargo doc --all --all-features --no-deps --document-private-items --features rkyv/size_32
 
   test:
     name: Test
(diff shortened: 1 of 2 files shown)


MarijnS95:
@bitshifter what's your preference to set these "required" features on dependent crates to have the least friction while developing `glam`? We've put it in `[dev-dependencies]` so that `--all-features` (as long as it is used with e.g. `--all-targets`), but that obviously isn't taken into account for `cargo doc`.

bitshifter:
I suspect I left default features enabled on rkyv because I ran into this problem. Ideally default features would be disabled for all dependencies. I wonder if there is something rkyv could do to make this easier to deal with.

MarijnS95:
@bitshifter Unfortunately I don't think `cargo` really allows for something better... except `cfg`s which can be controlled globally and don't require this mess as long as there's a sensible default somewhere.

bitshifter:
I guess I can live with rkyv being a dev-dependency. Adding `--features "rkyv/size_32"` to the doc build should fix it I think, at least that worked locally.

MarijnS95:
@bitshifter that works indeed. I'm not too keen on having it hardcoded there, but it seems to be the only solution :ok_hand:

PR #419, opened 2023-07-27
Implement div_euclid, rem_euclid

SpecificProtagonist:
Fixes #417.

I didn't find a way to vectorize this any any of the used instruction sets, but perhaps I'm just not clever enough (for floats, euclidean div is essentially a normal div followed by rounding away from the divisor).

### DIFF
--- a/codegen/templates/vec.rs.tera
+++ b/codegen/templates/vec.rs.tera
@@ -1080,7 +1080,7 @@ impl {{ self_t }} {
     ///
     /// # Panics
     /// This function will panic if any `rhs` element is 0 or the division results in overflow.
-    {% endif %}
+    {%- endif %}
     #[inline]
     pub fn div_euclid(self, rhs: Self) -> Self {
         Self::new(
@@ -1099,7 +1099,7 @@ impl {{ self_t }} {
     ///
     /// # Panics
     /// This function will panic if any `rhs` element is 0 or the division results in overflow.
-    {% endif %}
+    {%- endif %}
     ///
     /// [Euclidean division]: {{scalar_t}}::rem_euclid
     #[inline]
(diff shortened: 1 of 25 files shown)


codegen/templates/vec.rs.tera
```diff
@@ -1074,6 +1074,48 @@ impl {{ self_t }} {
...
+    /// # Panics
+    /// This function will panic if any `rhs` element is 0 or the division results in overflow.
+    {% endif %}
+    ///
+    /// [Euclidean division]: {{scalar_t}}::rem_euclid
+    {% if is_float %} #[cfg(not(feature = "libm"))] {% endif %}
+    #[inline]
+    pub fn rem_euclid(self, rhs: Self) -> Self {
+        let mut out = Self::default();
+        for i in 0..{{ dim }} {
```
bitshifter, line 1114: It is preferable to do the for in codegen so the generated code is already unrolled. There are many examples of this in the template, e.g. something like would probably do the job:
```
        Self::new(
            {% for c in components %}
                math::rem_euclid(self.{{ c }}, rhs[{% loop.index0 %}),
            {%- endfor %}
        )
```

codegen/templates/vec.rs.tera
```diff
@@ -1074,6 +1074,48 @@ impl {{ self_t }} {
     pub fn distance_squared(self, rhs: Self) -> {{ scalar_t }} {
         (self - rhs).length_squared()
     }
+
+    /// Returns the element-wise quotient of [Euclidean division] of `self` by `rhs`.
+    ///
+    {% if is_float%}
+    {% if not is_scalar %}
```
bitshifter, line 1081: If you look at the generated output from this there are some unwanted newlines, you can remove those by adding a `-` to `{%` or `%}`, e.g. `{%-`.

bitshifter (changes requested):

bitshifter (changes requested):
Mostly there, a  few more empty lines in generated code to sort out.

There are `rem_euclid` and `div_euclid` methods on `u32` and `u64` in Rust. Do you want to add those? I don't really mind either way, but the test code would need to be moved or duplicated to test unsigned types if you do add them.

bitshifter:
Thanks!

PR #421, opened 2023-07-31
Quaternion rotation fix and minor documentation fix

Aztro-dev:
Previously, quaternions were limited from `-PI / 2.0` to `PI / 2.0`, but that is now fixed and quaternions can now use all angles. Code was translated from Java to Rust using [this link](http://www.euclideanspace.com/maths/geometry/rotations/conversions/quaternionToEuler/)

I also fixed a minor documentation error where the word `used` was used twice on accident.

### DIFF
--- a/codegen/templates/quat.rs.tera
+++ b/codegen/templates/quat.rs.tera
@@ -31,7 +31,7 @@
 
 use crate::{
     {{ scalar_t }}::math,
-    euler::{EulerFromQuaternion, EulerRot, EulerToQuaternion},
+    euler::{EulerRot, EulerToQuaternion},
     {% if scalar_t == "f32" %}
         DQuat, Mat3, Mat3A, Mat4, Vec2, Vec3, Vec3A, Vec4,
     {% elif scalar_t == "f64" %}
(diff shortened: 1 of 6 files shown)


bitshifter:
This looks like it removes the functionality of specifying what order the Euler angles should be in?

n3vu0r:
You might also wanna have a look at the implementation of [nalgebra](https://docs.rs/nalgebra/latest/nalgebra/geometry/type.Rotation3.html#method.euler_angles_ordered). It's based on S2CID [123059499](https://api.semanticscholar.org/CorpusID:123059499).

Aztro-dev:
I'm sorry if I'm new to the codebase/concept, but what is the point of an `EulerRot`? To me, it just seemed like it moved around where the X, Y, and Z values where in the result. I can add `EulerRot`s to the pull request if you need.

bitshifter:
Rotations using Euler angles give different results depending on the order that the angles are applied in, EulerRot allowed users to specify what ordering they were using, glam does not make an assumption about a particular order being used.

I'm going to close this for now. I'd suggest opening an issue describing the problem that you are trying to solve, e.g. what you expect to happen and what is actually happening so that I can better understand the issue and assess any future PR.

PR #423, opened 2023-08-17
Correct the truncate documentation on 4D types

Indra-db:
the 4D types truncate mentions transforming it into a 2D type, while it should be 3D. This change fixes that.


bitshifter:
Most of glam is generated, you will need to update the template then generate the code, see
https://github.com/bitshifter/glam-rs/blob/main/codegen/README.md

bitshifter:
I've update the template and fixed this in main.

PR #425, opened 2023-08-30
Implement `MulAssign` for affine types

terrybrash:


bitshifter (commented):
Thanks! Would you be able to add tests for mul_assign as well? You can probably just copy the mul tests.

PR #427, opened 2023-09-02
Fix singularities in `Quat::to_euler()`

oisyn:
`Quat::to_euler()` has issues when the second rotation is (near) `PI/2` or `-PI/2` radians, causing the first and third angles to be incorrectly calculated.

For example:
```
    let q = Quat::from_euler(EulerRot::XYZ, 1.0, PI / 2.0, 0.0);
    let e = q.to_euler(EulerRot::XYZ);
    let q2 = Quat::from_euler(EulerRot::XYZ, e.0, e.1, e.2);
    println!("{:?}", q);
    println!("{:?}", e);
    println!("{:?}", q2);
```
Outputs:
```
Quat(0.33900508, 0.6205446, 0.33900508, 0.6205446)
(0.0, 1.5707964, 0.0)
Quat(0.0, 0.7071068, 0.0, 0.7071068)
```

This PR addresses this issue for all orders of rotation. It favors angles on the first axis over the third axis (e.g., a quaternion constructed from euler angles `(a, PI/2, b)` will convert back to either `(a+b, PI/2, 0)` or `(a-b, PI/2, 0)`, depending on the requested order). In any way, converting a quaternion to euler angles and back to a quaternion will always yield the original quaternion, barring any rounding errors.

Unfortunately, the fix requires a common prerequisite calculation for all three individual axes, so to avoid doing the same work multiple times I've removed the default implementation of `convert_quat()` and moved it to an implementation that basically does all the work that `first()`, `second()` and `third()` are doing, causing this code to be duplicated.

As far as the actual math is concerned, this is where the magic happens:
```
let sine_theta = self.sine_theta(q);
let second = math::asin(sine_theta);

if math::abs(sine_theta) > 0.99999 {
	let scale = 2.0 * math::signum(sine_theta);

	return match self {
		ZYX => (scale * math::atan2(-q.x, q.w), second, 0.0),
		ZXY => (scale * math::atan2(q.y, q.w), second, 0.0),
		YXZ => (scale * math::atan2(-q.z, q.w), second, 0.0),
		YZX => (scale * math::atan2(q.x, q.w), second, 0.0),
		XYZ => (scale * math::atan2(q.z, q.w), second, 0.0),
		XZY => (scale * math::atan2(-q.y, q.w), second, 0.0),
	};
}
```
So when the second rotation, `theta`, is `PI/2` or `-PI/2`, `sin(theta)` will be 1 or -1 respectively, and we can calculate the first rotation based on the table in this code.

### DIFF
--- a/src/euler.rs
+++ b/src/euler.rs
@@ -46,9 +46,10 @@ pub(crate) trait EulerFromQuaternion<Q: Copy>: Sized + Copy {
     fn third(self, q: Q) -> Self::Output;
 
     /// Compute all angles of a rotation in the notation order
-    fn convert_quat(self, q: Q) -> (Self::Output, Self::Output, Self::Output) {
-        (self.first(q), self.second(q), self.third(q))
-    }
+    fn convert_quat(self, q: Q) -> (Self::Output, Self::Output, Self::Output);
+
+    #[doc(hidden)]
+    fn sine_theta(self, q: Q) -> Self::Output;
 }
 
 /// Conversion from euler angles to quaternion.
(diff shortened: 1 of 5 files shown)


bitshifter (commented):
Thanks! Could you add the failing cases that this is fixing to the unit tests?

oisyn:
Done! It now tests for the full -180..=180 range on all axes 🙂.

Had to rewrite the tests a bit. It used to test that `to_euler()` returned the same values as given to `from_euler()`, but that's incorrect as those can be ambiguous. The same goes for quaternions: `-q` represents the same rotation as `q`, and `(x, y, z, 0)` for any value of `x`, `y` and `z` represents no rotation at all, so I had to make sure these all compare equal in the test.

oisyn:
Alright, they all seem to pass now 🚀

bitshifter:

PR #429, opened 2023-09-09
Implement wrapping/saturating integer arithmetic methods

### LINKED ISSUE #428: Adding wrapping/saturating methods on integer vectors
`UVec` and `IVec` are already useful for doing math because they implement all the math operations on themselves, but every time explicit wrapping/saturating operations are needed, all the operation needs to be applied to every element manually.

To overcome this, I would propose a new set of methods on integer vectors that mimic the behavior of their primitive type:
```rust
fn wrapping_add(self, rhs: Self) -> Self;
fn wrapping_sub(self, rhs: Self) -> Self;
fn wrapping_mul(self, rhs: Self) -> Self;
fn wrapping_div(self, rhs: Self) -> Self;

fn saturating_add(self, rhs: Self) -> Self;
fn saturating_sub(self, rhs: Self) -> Self;
fn saturating_mul(self, rhs: Self) -> Self;
fn saturating_div(self, rhs: Self) -> Self;
```

where the method calls the primitive equivalent on every element in the vector.

MrGunflame:
Closes #428.

### DIFF
--- a/codegen/templates/vec.rs.tera
+++ b/codegen/templates/vec.rs.tera
@@ -1727,6 +1727,112 @@ impl {{ self_t }} {
     }
     {% endif %}
 {% endif %}
+
+{% if not is_float %}
+    /// Returns a vector containing the wrapping addition of `self` and `rhs`.
+    ///
+    /// In other words this computes `[self.x.wrapping_add(rhs.x), self.y.wrapping_add(rhs.y), ..]`.
+    #[inline]
+    #[must_use]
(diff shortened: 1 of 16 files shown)


bitshifter:
Thanks! Could you add some tests for these new methods to this PR as well.

MrGunflame:
Added some basic tests, I'm not exactly sure in what form you would like tests for those.

bitshifter:
Thanks. Basic tests are fine, even calling code is something, if I refactor and delete/break something the test should hopefully catch it.

PR #432, opened 2023-09-23
Add `to_angle` method to 2D vectors

devil-ira:


tests/vec2.rs
```diff
@@ -911,6 +911,21 @@ macro_rules! impl_vec2_float_tests {
                 $vec2::new(0.0, -1.0)
             );
         });
+
+        glam_test!(test_to_angle, {
```
bitshifter, line 915: It would be good to add some tests that take an angle, create a vector with `from_angle` then make sure the same angle is returned by `to_angle` to make sure that they are behaving consistently.

bitshifter (changes requested):

PR #435, opened 2023-10-02
Add trait for extending Floating-Point Types with additional methods

### LINKED ISSUE #319: lerp over scalar types?
Hello!

I'm using glam via bevy. Ran into a situation where I need to lerp over a single f32 value i.e. non-vector. I couldn't find a function for it in glam, so I would either need use Vec2::lerp and de-structure the single value out of it, or write my own boilerplate code.

Is this even in the scope of this project? Or should this be implemented as utils in bevy?

devil-ira:
Closes #319

### DIFF
--- a/src/float.rs
+++ b/src/float.rs
@@ -1,4 +1,4 @@
-/// A trait for extending [`f32`] and [`f64`] with extra methods.
+/// A trait for extending [`prim@f32`] and [`prim@f64`] with extra methods.
 pub trait Float {
     /// Performs a linear interpolation between `self` and `rhs` based on the value `t`.
     ///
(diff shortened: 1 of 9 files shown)


src/float.rs
```diff
@@ -0,0 +1,28 @@
+/// A trait for extending [`prim@f32`] and [`prim@f64`] with extra methods.
+pub trait Float {
```
bitshifter, line 2: It might be worth calling this `FloatEx` or something. Mostly because there is `num_traits::Float` already which provides the same interface as native `f32`/`f64` in Rust where as this is adding additional functionality. So I'm worried about clashing with the `num_traits` trait and also providing different functionality from what people might expect from a `Float` trait.

bitshifter (changes requested):
I just had one concern around the naming of the `Float` trait.

src/float.rs
devil-ira: I had the same concern but then promptly forgot about it, oops! I totally agree.

I've gone with `FloatExt` for the name as that's the suffix I've used when adding "extension" traits within Bevy. If that's alright.

PR #439, opened 2023-11-23
add I16Vec and U16Vec

kawogi:
In response to https://github.com/bitshifter/glam-rs/discussions/260 I gave it a try to add the shorter vector types (`I16Vec_` and `U16Vec_`).

I still need to expand some of the test cases, but `vec2` is complete and looks good already.

I wasn't able to test all targets (esp. cuda) and I had to guess the alignments for cuda. Also there might be other places where the alignment isn't correct due to wrong assumptions. Somebody please have an eye on that.

Otherwise I added a few missing `From`/`TryFrom` conversions.

Feedback is welcome, so I know whether to invest more time in the remaining tests.

Edit: I just found the test runner for the features. Will do some fixing …


kawogi:
`build_and_test_features.sh` shows all green now.

Only the tests for `vec3` and `vec4` are missing and I'm still unsure whether the CUDA alignments are correct.

bitshifter (approved):
Nice, thanks for all of your work on this!

kawogi:
Thanks for the crate!

Are the test cases for vec3/vec4 still required?

bitshifter:
Ah yes that's a good point, coverage did drop a bit with this PR (see https://coveralls.io/builds/64166568), I forgot to look. It would be good if you could add those missing tests.

kawogi:
I actually added the missing test cases (and some more) on my fork (https://github.com/kawogi/glam-rs/tree/main/src) but after wasting over an hour I couldn't figure out how to make a new PR from that state! I'm totally lost with GIT(Hub) right now. Might try tomorrow again. Any help is appreciated.

bitshifter:
When a PR is merged the default behaviour in github is to squash the changes (you can see the squashed commit here https://github.com/bitshifter/glam-rs/commit/03c2233ebbcd3b54ad84668050fa9f5b15a162f2). This means that the history gets changed, so if you continue to work on the same branch your original changes are on then git will think they aren't related.

When making a PR you should first create a local branch, make your changes, commit the branch, push it to github and make the PR from that. Once the PR is merged you can delete the branch. Since your changes are on main, it can be a bit tricky, you might be able to rename your main branch, then checkout the current main from upstream glam to a new main so your main is correct. Then create a new branch and either file copy or git cherry pick the changes you made from the old main. If this is too much hassle let me know and I can try cherry pick the changes you made from your fork.

kawogi:
Thanks a lot for your explanation and I think I understood the problem. For now, I cherry-picked the changes on my own into a new fork but will remember your suggestions for the next time.

Follow-up: https://github.com/bitshifter/glam-rs/pull/440

PR #440, opened 2023-11-26
add more i16/u16 tests; fix cuda align for i16/u16

kawogi:
This is a follow-up for https://github.com/bitshifter/glam-rs/pull/439, adding the missing test cases (and some more).

Also the CUDA-alignment has been fixed (to my knowledge).

Test-coverage should at least be back to normal. API is backwards-compatible (no version-bump required IMO).

### DIFF
--- a/codegen/templates/vec.rs.tera
+++ b/codegen/templates/vec.rs.tera
@@ -91,10 +91,22 @@
     {% set try_from_types = ["I16Vec" ~ dim, "IVec" ~ dim, "I64Vec" ~ dim] %}
 {% endif %}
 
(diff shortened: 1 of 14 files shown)


bitshifter:
This is the alignment table I've been using, https://docs.nvidia.com/cuda/archive/11.7.0/cuda-c-programming-guide/index.html#vector-types__alignment-requirements-in-device-code. I think what you have looks about right.

Thanks for all your work here!

kawogi:
Thanks for checking. I'm happy to give something back and your crate was super easy to work with.

PR #449, opened 2023-12-17
added-must-use

sanbox-irl:
Added `#[must_use]` to those methods which look like `(&*self ...) -> Self` (basically methods which take a self by non-mutable ref or by value, and which return a Self). However, I did not add such bounds to trait impls whose trait definition already had such a `must_use` on it (all the core::ops and all the From impls). Along those notes, I added `[#must_use]` to `FloatExt`'s trait definition.

On this (m1 mac), most of the tests except for those in "euler::quat::test_" did not work on main currently and after this, continue to not work. It appears that they work fine on the Github Actions runner, so I suspect that this PR should be fine, since this should not result in any different code generated.

I tried to check all the Tera files as appropriate, and I think I got them all, but even if I didn't, I would be comfortable with this PR merging, as it's just a continuation of efforts to match the std's new `must_use` style.

### DIFF
--- a/codegen/templates/mat.rs.tera
+++ b/codegen/templates/mat.rs.tera
@@ -1895,6 +1895,7 @@ impl {{ self_t }} {
 {% endif %}
 
     /// Multiplies two {{ nxn }} matrices.
+    #[must_use]
     #[inline]
     pub fn mul_mat{{ dim }}(&self, rhs: &Self) -> Self {
         {% if self_t == "Mat2" and is_sse2 %}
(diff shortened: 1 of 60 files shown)


bitshifter:
Hmm, it looks like the codegen lint failed. It looks like you ran the code generator, what version of rustc are you on?

sanbox-irl:
`1.74.0` -- do i need to be on a specific RV?

bitshifter:
That should be ok. Just checking, you ran `cargo run -p codegen`? I might need to check it out locally and try work out what's up

sanbox-irl:
ah @bitshifter going to check but my fault -- ran `cargo run -p codegen` and then updated the tera files (standardized the must_use vs. inline order) and forgot to return

bitshifter:
Thanks!

PR #455, opened 2023-12-26
In some build configs, BVec4 was an unused import.

waywardmonkeys:
The uses of this were protected by only `scalar_math` so only protect the import by the same.


bitshifter:
Thanks!

Issue #456, opened 2024-01-02
Make `Mat[x]::new()` functions public


LegNeato:
The `Mat[x]::new()` functions are not public but they are for `Vec`s. Is there a reason we must use `from_cols` or the fields directly instead of calling `new()`? For reference, here is the private function for `Mat3`:

```rust
    #[allow(clippy::too_many_arguments)]
    #[inline(always)]
    #[must_use]
    const fn new(
        m00: f32,
        m01: f32,
        m02: f32,
        m10: f32,
        m11: f32,
        m12: f32,
        m20: f32,
        m21: f32,
        m22: f32,
    ) -> Self {
        Self {
            x_axis: Vec3::new(m00, m01, m02),
            y_axis: Vec3::new(m10, m11, m12),
            z_axis: Vec3::new(m20, m21, m22),
        }
    }
```

I put up a PR in case this is an oversight rather than an explicit choice.

bitshifter:
The purpose of `from_cols` is that it explicitly describes that the input data should be column vectors as opposed to row vectors, this is why raw `new` is not exposed as is.  If I were to expose a new method that takes values in column order I'd suggest calling it `from_cols_values` which would use similar naming to existing methods.

PR #457, opened 2024-01-02
Make `Mat[x]::new()` public.

LegNeato:
`Mat` now matches `Vec` as `Vec[x]::new()` is already public. Fixes https://github.com/bitshifter/glam-rs/issues/456.


bitshifter (changes requested):
As I mentioned in the related issue, I think this should follow glam's explicit naming scheme that other methods use, e.g. there are currently `from_cols`, `from_cols_array` and `from_cols_array_2d`, so perhaps `from_cols_values` would fit. If you think that's still a useful addition I'm happy to accept a PR.

LegNeato:
Great, will rework if I need it!

PR #464, opened 2024-01-22
made vec conversions const

sanbox-irl:
Added `const` to the `as_` conversions for Vec. This might also be appropriate in other places, but this must is what I had time and familiarity enough for!

### DIFF
--- a/codegen/templates/vec.rs.tera
+++ b/codegen/templates/vec.rs.tera
@@ -1726,28 +1726,28 @@ impl {{ self_t }} {
     /// Casts all elements of `self` to `f32`.
     #[inline]
     #[must_use]
-    pub fn as_vec2(&self) -> crate::Vec2 {
+    pub const fn as_vec2(&self) -> crate::Vec2 {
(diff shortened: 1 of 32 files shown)


sanbox-irl:
hmm the errors the CI is getting i am not getting on my local!

bitshifter:
Are you on an arm chip? Platforms that use SIMD like x86_64 use deref coercion to access data members, this is not const fn.

sanbox-irl:
I am indeed! I targeted amd64 intel for macos and can see the errors locally now -- I suspect that the utility of this PR is not worth the amount of maintenance burden it would introduce since we'd have to branch in the tera scripts on sse2 for just the const. Is there a way to do basically do:

```
pub NO_SSE2_CONST fn /* everything else */
```
is such a way that's easy to maintain?

bitshifter:
Hey, so as a rule I try to make the API consistent no matter what architecture you are building for. Otherwise, say in this example if you were using the `const fn as_vec4` in a `const` context your code might compile on arm but not x86, which is not ideal. Given that, I'm going to close this PR.

PR #465, opened 2024-01-25
Element-wise sum and product for vectors. (#462)

Ababwa:


Ababwa:
Compiler can't find `Which` in `std::simd`. Not sure what to do here. https://doc.rust-lang.org/std/simd/enum.Which.html

bitshifter:
The docs might be out of date. I think they got rid of First and Second, numbers should be the same as what wasm uses now.

Ababwa:
Made some revisions. Should be good now.

Ababwa:
Closing to change fork branch.

PR #471, opened 2024-02-19
Element-wise absolute value for matrices

NLincoln:
Pretty much exactly as the title says: add an `abs()` method to matrix types, that calls `abs()` on each column vector.

### DIFF
--- a/codegen/templates/mat.rs.tera
+++ b/codegen/templates/mat.rs.tera
@@ -2086,6 +2086,17 @@ impl {{ self_t }} {
         {% endfor %}
     }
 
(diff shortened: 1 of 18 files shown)


bitshifter:
I'm curious, what is your use case for this PR?

bitshifter:
All good, I'm just interested in how things in glam get used. I guess you are specifically implementing this bit https://pbr-book.org/4ed/Shapes/Managing_Rounding_Error#eq:transform-point-error?

bitshifter (approved):

PR #473, opened 2024-02-23
Fix FloatExt::lerp's documentation

ekropotin:
FloatExt::lerp's documentation mentions a value `t` in the method signature that doesn't exist. It probably means `s`.
Fixes #472


bitshifter:
Thanks!

PR #484, opened 2024-03-05
Adds support for a Vec::normalize_or(fallback) method

TheComet:
See https://github.com/bitshifter/glam-rs/discussions/483


bitshifter:
Just needs a `cargo fmt` by the looks of it.

bitshifter:
Thanks!

PR #485, opened 2024-03-06
fix typo: addres -> address

ameknite:


bitshifter:
Thanks, but I think `addres` is short for `add result` here, matching `subres` which is short for `subtract result`.

ameknite:
oh, i didn't realize lol

Issue #477, opened 2024-02-28
Feature request: add `towards` for vectors


chompaa:
I'm not sure if this fits within the scope of the library, but I think interpolating in this fashion could be useful.

Rough implementation:

```rust
fn towards(&self, rhs: Self, delta: f32) -> Self {
  let a = rhs - self;
  let len = a.length();
  
  if len <= delta || len <= 1e-4 {
    return rhs;
  }
    
  self + a / len * delta
}
```

I saw that this was somewhat discussed briefly in #265, but am unsure if it's wanted.

bitshifter:
Note earlier discussion here #265

PR #492, opened 2024-03-11
Add `move_towards` for vectors

chompaa:
Adds a `move_towards` function for vectors based on #477 and #265, with tests.

### DIFF
--- a/codegen/templates/vec.rs.tera
+++ b/codegen/templates/vec.rs.tera
@@ -1615,7 +1615,7 @@ impl {{ self_t }} {
     /// `self.distance(rhs)`, the result will be equal to `rhs`. Will not go past `rhs`.
     #[inline]
     #[must_use]
-    pub fn towards(&self, rhs: Self, d: {{ scalar_t }}) -> Self {
+    pub fn move_towards(&self, rhs: Self, d: {{ scalar_t }}) -> Self {
         let a = rhs - *self;
         let len = a.length();
         if len <= d || len <= 1e-4 {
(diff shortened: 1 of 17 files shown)


codegen/templates/vec.rs.tera
```diff
@@ -1609,6 +1609,21 @@ impl {{ self_t }} {
         self + ((rhs - self) * s)
     }
 
+    /// Moves towards `rhs` based on the value `d`.
```
bitshifter, line 1612: I think it would be worth wrapping this in `{% if dim < 4 %}` because I don't know that it would be useful on the Vec4.

bitshifter (changes requested):

codegen/templates/vec.rs.tera
bitshifter: On second thoughts, leave this as is, I already have a bunch of other methods on Vec4 that are mathematically valid but probably not useful, so this won't be out of place at least.

codegen/templates/vec.rs.tera
```diff
@@ -1609,6 +1609,21 @@ impl {{ self_t }} {
...
+    ///
+    /// When `d` is `0.0`, the result will be equal to `self`. When `d` is equal to 
+    /// `self.distance(rhs)`, the result will be equal to `rhs`. Will not go past `rhs`.
+    #[inline]
+    #[must_use]
+    pub fn towards(&self, rhs: Self, d: {{ scalar_t }}) -> Self {
```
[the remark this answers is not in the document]
chompaa: Agreed, I've changed the name accordingly.

bitshifter (approved):

PR #498, opened 2024-03-20
ci: Don't use actions-rs actions

waywardmonkeys:
These haven't been maintained in some time, so just do the steps directly as other jobs do.


bitshifter:
Thanks!

PR #505, opened 2024-04-19
Fix clippy::legacy_numeric_constants lints.

waywardmonkeys:
The `std::f64::EPSILON` and similar constants are being deprecated and it is encouraged to use the associated constants on `f32` and `f64` instead.


waywardmonkeys:
This is a newer lint that is present in the nightly clippy.

bitshifter:
Thanks!

PR #507, opened 2024-04-30
Unreal ABI Compatibility

herohiralal:
Updates:
- Added feature `unreal-abi-compat`.
  - Corrects the alignments for `DQuat` and `DVec4`.
- Added const `UNREAL_EULER_ROT`.
  - Exposes the default Unreal Engine rotation order.
- Added `Rotator` and `DRotator`.
  - These correspond to `FRotator3f`/`FRotator3d`.
  - Provide degree-based euler-angle proxy.
  - Contains euler/quaternion cast operators.
  - Wrapped in feature.

Minor changes only.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "glam"
-version = "0.27.0" # remember to update html_root_url
+version = "0.27.0"                                                       # remember to update html_root_url
 edition = "2021"
 authors = ["Cameron Hart <cameron.hart@gmail.com>"]
 description = "A simple and fast 3D math library for games and graphics"
@@ -41,6 +41,9 @@ fast-math = []
 # experimental nightly portable-simd support
 core-simd = []
 
+# Unreal Engine ABI Compatibility, for interoperating with Rust
+unreal-abi-compat = []
+
 [dependencies]
 approx = { version = "0.5", optional = true, default-features = false }
 bytemuck = { version = "1.9", optional = true, default-features = false }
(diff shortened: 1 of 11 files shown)


bitshifter:
For this kind of thing it's best to start a [discussion](https://github.com/bitshifter/glam-rs/discussions) (see https://github.com/bitshifter/glam-rs?tab=readme-ov-file#contribution).

I don't think that a lot of this makes sense for glam. It may be preferable to maintain a separate crate for interop with Unreal.

As far as I can tell in UE5 FVector3f and FVector4f are not 16 byte aligned, they, see https://github.com/EpicGames/UnrealEngine/blob/release/Engine/Source/Runtime/Core/Public/Math/MathFwd.h declarations of `FVector`, `FVector3d` and `FVector3f` which are specializations of `TVector`, there is no alignment specified as far as I can tell. There are types like `VectorRegister4Float` but they are just typedefs to register types for SSE, Neon, etc, not general purpose Vector3 types, so I don't think it makes sense to add alignment to glam types like this.

I also don't want to maintain a copy of Unreal's Rotator type, I don't think it makes sense in the context of glam.

I'm going to close this PR but feel free to start a discussion if you want to explain what you are wanting further.

herohiralal:
[FQuat4f/FQuat4d](https://github.com/EpicGames/UnrealEngine/blob/16dc333db3d6439c7f2886cf89db8907846c0e8a/Engine/Source/Runtime/Core/Public/Math/Quat.h#L37) and [FVector4f/FVector4d](https://github.com/EpicGames/UnrealEngine/blob/16dc333db3d6439c7f2886cf89db8907846c0e8a/Engine/Source/Runtime/Core/Public/Math/Vector4.h#L25) both are. Hyperlinks are to the lines setting the alignment.

Using it in a personal project regardless of the addition to the main crate, but thought it might be helpful to others too. And might help ease up some amount of pains involved in using Rust for gamedev (having to start in feature-scarce engines or start from scratch), by instead being more accessibly integrateable into a more popular engine.

Not going to pursue integration further, and so just posting this minor info in this context for visibility.

bitshifter:
Thanks for the additional context, and the links. Visual Assist never seems to find the right thing for Unreal math types.

Optional alignment is a hassle to maintain which is a big reason I wouldn't want to take this until it's mature and there's a some demand.

PR #509, opened 2024-05-17
clippy: Fix `let_and_return` lint in neon.

waywardmonkeys:


bitshifter:
This one is generated so you'll need to update the template and rerun codegen, or I can do it if you prefer.

bitshifter:
Looking at the template it might be easier to add `It might be easier to add `#[allow(clippy::let_and_return)]` here.

waywardmonkeys:
I just updated the template and thought about the same thing ... but tried this as well. What do you think?

bitshifter:
That works as well :)

waywardmonkeys:
I had also tried just doing `unsafe { ... } & 0x7` but that didn't looked bad to me in the generated re-formatted code.

waywardmonkeys:
And the reason this didn't get picked up by CI is that clippy is just run on the ubuntu build, not all platforms / configs.

bitshifter:
Ah, I was wondering. I should do something about that.

PR #510, opened 2024-05-18
clippy: Fix `semicolon_if_nothing_returned` lints.

waywardmonkeys:


bitshifter:
Hmm, I tried running clippy on macos as well but it didn't catch this, what version of Rust are you on?

https://github.com/bitshifter/glam-rs/actions/runs/9137851611/job/25128322722?pr=511

waywardmonkeys:
I enabled it manually!

We could probably add a `[lints]` configuration to the `Cargo.toml` to add this and maybe some others...

PR #511, opened 2024-05-18
Run clippy on ubuntu and macos for better coverage.

bitshifter:


bitshifter:
Might not need this after all.

PR #517, opened 2024-05-27
Fix vector clamp length methods to not accept negative values

### LINKED ISSUE #515: Vec2 clamp_length alters vector when length is within specified range
The `clamp_length` method in the `Vec2` incorrectly alters the vector components when the vector's length is within the specified `min` and `max` range. This behavior deviates from the expected functionality where the vector should remain unchanged if its length falls within the specified bounds.

```rust
let vec = Vec2::new(3., 4.);
let new_vec = vec.clamp_length(-10., 10.);
println!("{}", new_vec);
//the new_vec is [-6, -8]
```

While it is possible that this behavior is intentional in the current implementation, it seems counterintuitive for the method to alter the vector when its length is already within the specified range. Clarification on whether this is the expected behavior would be appreciated.

semihbkgr:
fixes #515

### DIFF
--- a/codegen/templates/vec.rs.tera
+++ b/codegen/templates/vec.rs.tera
@@ -1870,7 +1870,7 @@ impl {{ self_t }} {
     #[inline]
     #[must_use]
     pub fn clamp_length(self, min: {{ scalar_t }}, max: {{ scalar_t }}) -> Self {
-        glam_assert!(0 <= min);
+        glam_assert!(0 as {{ scalar_t }} <= min);
         glam_assert!(min <= max);
         let length_sq = self.length_squared();
         if length_sq < min * min {
@@ -1890,7 +1890,7 @@ impl {{ self_t }} {
     #[inline]
     #[must_use]
     pub fn clamp_length_max(self, max: {{ scalar_t }}) -> Self {
-        glam_assert!(0 <= max); 
+        glam_assert!(0 as {{ scalar_t }} <= max); 
         let length_sq = self.length_squared();
         if length_sq > max * max {
             max * (self / math::sqrt(length_sq))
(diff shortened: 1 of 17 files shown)


bitshifter:
That looks good to me. Note that most code in glam is generated, see https://github.com/bitshifter/glam-rs/blob/main/codegen%2FREADME.md for how to run codegen.

Could you add a test for the problem you are fixing as well?

bitshifter:
Sorry, I had some more thoughts on this which I've posted on the original issue.

bitshifter:
GitHub is saying there are merge conflicts, not sure why.

You'd need to pull latest. Usually I rebase on top of latest and force push to the PR branch.

bitshifter:
Nevermind, I think GitHub mobile was confused

semihbkgr:
I got the same conflit on my local fork. To resolve, I just merged the latest `main` branch into this branch.

codegen/templates/vec.rs.tera
```diff
@@ -1862,14 +1862,15 @@ impl {{ self_t }} {
...
+    /// Returns a vector with a length no less than `min` and no more than `max`.
     ///
     /// # Panics
     ///
-    /// Will panic if `min` is greater than `max` when `glam_assert` is enabled.
+    /// Will panic if `min` is greater than `max`, or if either `min` or `max` is negative, when `glam_assert` is enabled.
     #[inline]
     #[must_use]
     pub fn clamp_length(self, min: {{ scalar_t }}, max: {{ scalar_t }}) -> Self {
+        glam_assert!(0 as {{ scalar_t }} <= min);
```
bitshifter, line 1873: This method is only used in float vectors so it can be 0.0 without the need for a cast

bitshifter (changes requested):

codegen/templates/vec.rs.tera
semihbkgr: ohh, sure 👍

bitshifter (approved):

bitshifter:
Thanks for reporting this and the PR!

PR #528, opened 2024-06-20
Added bvec constructors

vallentin:
I noticed that all `BVec*`s are missing the freestanding `fn *vec*()` constructors.

### DIFF
--- a/codegen/templates/vec_mask.rs.tera
+++ b/codegen/templates/vec_mask.rs.tera
@@ -51,7 +51,18 @@ union UnionCast {
 }
 {% endif %}
 
-{%- if is_scalar and is_bool %}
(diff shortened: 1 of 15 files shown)


vallentin:
The lint was triggered by a missing cfg attr for `scalar-math`. Fixed it and rebase.

bitshifter (changes requested):
Looks good, I have one small request, could you add some test functions? While these are trivial functions it's useful to have tests for code coverage (to see what still needs testing) and also if I ever refactor to make sure I don't miss some functionality.

If you look in tests/vec*.rs there are tests that start with `test_mask_`, you could either add them to an existing test or add a new one that does something like `assert_eq!($mask::new(false, false), $bvec2(false, false))`

One catch is you would need to add `$bvec2` to the macro signature and pass it through.

Issue #441, opened 2023-11-30
Why is there no `Mat4::frustum_{lh,rh{,_gl}}` ?


Des-Nerger:
I thought that, not only `Mat4::perspective_{lh,rh{,_gl}}` requires more computation involving the nasty `math::tan`, it also doesn't allow us to construct matrices for stereoscopic projection: https://paulbourke.net/stereographics/stereorender

One could omit `perspective` (if the minimalistic approach were pursued), but `frustum` is an essential and fundamental thing to have for such a math library.

bitshifter:
I am not understanding exactly what is being asked for here, could you explain exactly what feature you would like to see?

bitshifter:
A more detailed explanation from https://www.opengl.org/archives/resources/faq/technical/transformations.htm

> What are the pros and cons of using glFrustum() versus gluPerspective()? Why would I want to use one over the other?
>
>  glFrustum() and gluPerspective() both produce perspective projection matrices that you can use to transform from eye coordinate space to clip coordinate space. The primary difference between the two is that glFrustum() is more general and allows off-axis projections, while gluPerspective() only produces symmetrical (on-axis) projections. Indeed, you can use glFrustum() to implement gluPerspective(). However, aside from the layering of function calls that is a natural part of the GLU interface, there is no performance advantage to using matrices generated by glFrustum() over gluPerspective().
>
>  Since glFrustum() is more general than gluPerspective(), you can use it in cases when gluPerspective() can't be used. Some examples include [projection shadows](https://www.opengl.org/archives/resources/faq/technical/lights.htm#ligh0130), tiled renderings, and stereo views.
>
>  Tiled rendering uses multiple off-axis projections to render different sections of a scene. The results are assembled into one large image array to produce the final image. This is often necessary when the desired dimensions of the final rendering exceed the OpenGL implementation's maximum viewport size.
>
>  In a stereo view, two renderings of the same scene are done with the view location slightly shifted. Since the view axis is right between the “eyes”, each view must use a slightly off-axis projection to either side to achieve correct visual results.

Des-Nerger:
> A more detailed explanation from https:\/\/www.opengl.org\/...

Yes. It's what I talked about. I haven't found an analogue of `glFrustrum` in `glam`, found only that of `gluPerspective`; which felt unjust. Of course one can construct `glFrustum` manually, but I thought `glam` is supposed to contain such conveniences by itself.

bitshifter:
[...]

PR #529, opened 2024-06-21
`Mat4::frustum_{lh,rh{,_gl}}`

valaphee:
Solves https://github.com/bitshifter/glam-rs/issues/441


bitshifter:
Thanks for taking this on!

It looks like clippy is failing because perspective is unused in the tests you've added, from your comment above it sounded like you were intending to use that value in the test though?

As for perspective being constructed from frustum, I don't have a strong opinion. Re-use is nice when it makes sense and a little perf hit isn't that big a deal IMO in these as I think they're not usually frequently called in hot code, but what might that look like?

bitshifter:
Superseded by #640

PR #532, opened 2024-06-27
fix(f32): is_normalized

5-23:


5-23:
image
XD

bitshifter:
Can you please provide a text description for this and update the tests to include the problem you are fixing

5-23:
> Can you please provide a text description for this and update the tests to include the problem you are fixing

This code eliminates the error that occurs when the length of vec2 is smaller than the minimum value of f32.

5-23:
While playing with bevy_xpbd, an error occurred, so I searched for it and estimated the cause.

bitshifter:
I don't really think this change makes sense, at least it is inefficient because the code now calculates length_squared twice, adds an if statement and it makes the check potentially too narrow, but I would like to better understand the problem you are trying to fix.

bitshifter:
Closing this, if you can provide some example code showing how is_normalized is doing something unexpected please open an issue.

PR #533, opened 2024-06-29
Fix documentation on supported architectures for SIMD

BD103:
This updates the crate documentation and `README.md` to account for all supported platforms for SIMD. I believe I got everything correct, but please fact-check it!

Have a great day :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ SIMD is supported on `x86`, `x86_64` and `wasm32` targets.
   `RUSTCFLAGS`.
 * To enable `simd128` on `wasm32` targets add `-C target-feature=+simd128` to
   `RUSTFLAGS`.
+* To enable `NEON` on `aarch64` targets add `-C target-feature=+neon` to `RUSTFLAGS`.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -62,6 +62,7 @@ SIMD is supported on `x86`, `x86_64` and `wasm32` targets.
   `RUSTCFLAGS`.
 * To enable `simd128` on `wasm32` targets add `-C target-feature=+simd128` to
   `RUSTFLAGS`.
+* To enable `NEON` on `aarch64` targets add `-C target-feature=+neon` to `RUSTFLAGS`.
```
bitshifter, line 65: This is a bit less clear to me on aarch64. For the platforms I tested on which were raspberry pi 4 and Mac M1,  NEON was enabled by default.  For now it's probably better to copy the x86_64 line and say it's on by default.

bitshifter (changes requested):

README.md
BD103: Got it! I searched online, and it looks like `NEON` is required for `aarch64` processors.

bitshifter:
Thanks for the PR!

PR #534, opened 2024-07-05
Make random quaternions uniformly random (via `Standard`)

mweatherley:
Previously, randomly sampled `Quat`s had bias because they were constructed using Euler rotation sequences. This PR makes them follow a uniform distribution. The algorithm used is the one described [here](https://web.archive.org/web/20180724074257/http://planning.cs.uiuc.edu/node198.html). 

This is essentially a follow-up from [this bevy_math PR](https://github.com/bevyengine/bevy/pull/12857), which affords users the ability to ergonomically obtain random directions and quaternions.

I believe that this implementation also conforms more closely with the philosophy of [Standard](https://docs.rs/rand/latest/rand/distributions/struct.Standard.html): since the space of quaternions is compact, it is quite natural for the "default" distribution over them to be uniform. 

Furthermore, if a user wants the old distribution, it is easily implemented using just `Rng::gen_range` and `Quat`'s constructors; the present implementation actually bundles relatively specialized knowledge into an interface, which abstracts difficulties away from users.

### DIFF
--- a/src/features/impl_rand.rs
+++ b/src/features/impl_rand.rs
@@ -84,10 +84,11 @@ macro_rules! impl_float_types {
         impl Distribution<$quat> for Standard {
             #[inline]
             fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> $quat {
-                let yaw = -PI + rng.gen::<$t>() * 2.0 * PI;
-                let pitch = -PI + rng.gen::<$t>() * 2.0 * PI;
-                let roll = -PI + rng.gen::<$t>() * 2.0 * PI;
-                $quat::from_euler(crate::EulerRot::YXZ, yaw, pitch, roll)
+                let u = rng.gen_range::<$t, _>(0.0..=1.0);
+                let (u1, u2) = ($t::sqrt(1.0 - u), $t::sqrt(u));
+                let (sin1, cos1) = rng.gen_range::<$t, _>(0.0..=TAU).sin_cos();
+                let (sin2, cos2) = rng.gen_range::<$t, _>(0.0..=TAU).sin_cos();
+                $quat::from_xyzw(u1 * sin1, u1 * cos1, u2 * sin2, u2 * cos2)
             }
         }
 
@@ -140,7 +141,7 @@ macro_rules! impl_float_types {
 
 mod f32 {
     use crate::{Mat2, Mat3, Mat4, Quat, Vec2, Vec3, Vec3A, Vec4};
-    use core::f32::consts::PI;
+    use core::f32::consts::TAU;
     use rand::{
         distributions::{Distribution, Standard},
         Rng,
@@ -169,7 +170,7 @@ mod f32 {
 
 mod f64 {
     use crate::{DMat2, DMat3, DMat4, DQuat, DVec2, DVec3, DVec4};
-    use core::f64::consts::PI;
+    use core::f64::consts::TAU;
     use rand::{
         distributions::{Distribution, Standard},
         Rng,
--- a/src/features/impl_rand.rs
+++ b/src/features/impl_rand.rs
@@ -84,11 +84,12 @@ macro_rules! impl_float_types {
         impl Distribution<$quat> for Standard {
             #[inline]
             fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> $quat {
-                let u = rng.gen_range::<$t, _>(0.0..=1.0);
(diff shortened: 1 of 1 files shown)


NthTensor:
Over at bevy we've had several users be confused when they encounter the current nonuniform version. We can't really fix this in our side, and to me this seems like a better default.

bitshifter (approved):

bitshifter:
You need to use `math::sin_cos` which wraps libm and std implementations.

ickk (changes requested):
I compared this PR to a few other approaches

The prng used was `rand::rngs::SmallRng`

```
rand_quat/existing      time:   [1.2494 ns 1.2528 ns 1.2561 ns]
rand_quat/subgroup      time:   [2.3529 ns 2.3573 ns 2.3618 ns]
rand_quat/sphere_inline time:   [1.8076 ns 1.8121 ns 1.8167 ns]
rand_quat/hatbox        time:   [1.2487 ns 1.2517 ns 1.2548 ns]
```

<details>
<summary>Code</summary>

```rust
#[inline]
pub fn existing<R: Rng>(rng: &mut R) -> Quat {
    rng.gen()
}

#[inline]
pub fn subgroup<R: Rng>(rng: &mut R) -> Quat {
    let (s1, c1) = rng.gen_range(0f32..TAU).sin_cos();
    let (s2, c2) = rng.gen_range(0f32..TAU).sin_cos();

    let x0 = rng.gen_range(0f32..1.);
    let (r1, r2) = (x0.sqrt(), (1. - x0).sqrt());

    Quat::from_xyzw(s1 * r1, c1 * r1, s2 * r2, c2 * r2)
}

#[inline]
pub fn sphere_inline<R: Rng>(rng: &mut R) -> Quat {
    let phi_cos = rng.gen_range(-1f32..=1.0);
    let phi_sin = phi_cos.acos().sin();
    let (theta_sin, theta_cos) = rng.gen_range(0f32..=TAU).sin_cos();

    let axis: Vec3 = Vec3::new(phi_sin * theta_cos, phi_sin * theta_sin, phi_cos);
    let angle = rng.gen_range(-TAU..=TAU);

    Quat::from_axis_angle(axis, angle)
}

#[inline]
pub fn hatbox<R: Rng>(rng: &mut R) -> Quat {
    let z = rng.gen_range(-1.0..=1.0);
    let (a_sin, a_cos) = rng.gen_range(-PI..=PI).sin_cos();
    let c = (1.0-z*z).sqrt();
    let x = a_sin*c;
    let y = a_cos*c;

    let axis = Vec3::new(x, y, z);
    let angle = rng.gen_range(-TAU..=TAU);

    Quat::from_axis_angle(axis, angle)
}
```
</details>

src/features/impl_rand.rs
```diff
@@ -84,10 +84,11 @@ macro_rules! impl_float_types {
...
             #[inline]
             fn sample<R: Rng + ?Sized>(&self, rng: &mut R) -> $quat {
-                let yaw = -PI + rng.gen::<$t>() * 2.0 * PI;
-                let pitch = -PI + rng.gen::<$t>() * 2.0 * PI;
-                let roll = -PI + rng.gen::<$t>() * 2.0 * PI;
-                $quat::from_euler(crate::EulerRot::YXZ, yaw, pitch, roll)
+                let u = rng.gen_range::<$t, _>(0.0..=1.0);
+                let (u1, u2) = ($t::sqrt(1.0 - u), $t::sqrt(u));
+                let (sin1, cos1) = rng.gen_range::<$t, _>(0.0..=TAU).sin_cos();
+                let (sin2, cos2) = rng.gen_range::<$t, _>(0.0..=TAU).sin_cos();
```
bitshifter, line 90: Regarding the test fail, these need to be changed to `math::sin_cos(rng.gen_range::<$t, _>(0.0..=TAU))`

`$t::sqrt` also needs to change to `math::sqrt`, you'll also need to add a `use crate::f32::math` or `use crate::f64::math` in the relevant `mod` block.

bitshifter:
I don't have a strong feeling about performance characteristics of this. It's something I added at the very beginning of glam to generate some data to run benchmarks on, it's not something I put any thought into at the time. It seems like there's still a bit of discussion happening on bevy discord, let me know what you settle on.

mweatherley:
Cool. I ended up switching to axis-angle randomization since I think it's simple to grok and its performance seems pretty good.

bitshifter:
Thanks for the PR!

PR #542, opened 2024-07-20
div_ceil for UVec

kornelski:
Convenience function for divison with rounding up. It's handy for things like `let num_tiles = world_size.div_ceil(tile_size);`

It's only for unsigned scalars, because Rust hasn't stabilized `div_ceil` for other types yet. I've copied `div_ceil` implementation to keep MSRV under 1.73.

### DIFF
--- a/codegen/templates/vec.rs.tera
+++ b/codegen/templates/vec.rs.tera
@@ -2584,6 +2584,19 @@ impl {{ self_t }} {
                 {%- endfor %}
             }
         }
+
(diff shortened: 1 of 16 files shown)


bitshifter:
Looks like it needs a rustfmt

kornelski:
Rustfmt's overly aggressive pedantism is such a waste of time… ;( The tool is so dumb that it can only automate breaking PRs, but can't automate fixing them.

PR #547, opened 2024-08-08
Generic mapping function for vectors

Aceeri:
Hi! Sorry I didn't make a discussion about this first, didn't read the guidelines until I already wrote this up, so figured I might as well post the PR.

This adds a generic mapping function that acts on all of the elements of a vector. Adding it to matrices as well might be useful, but my use cases for this has always just been vectors so far. I'm happy to add it to matrices if you think that'd be good for consistency's sake.

The use case, as an example, I had recently was taking any vector element in a normal below some `EPSILON` and setting it to 0.0, since the collision normal that was generated was a little bit finnicky. I don't see that use case alone as very useful in the library, but allowing per-element operations definitely is imo.

### DIFF
--- a/codegen/templates/vec.rs.tera
+++ b/codegen/templates/vec.rs.tera
@@ -411,6 +411,20 @@ impl {{ self_t }} {
         {% endif %}
     }
 
+    /// Returns a vector containing each element of `self` modified by a mapping function `f`.
+    #[inline]
+    #[must_use]
+    pub fn map<F>(self, f: F) -> Self
+    where
+        F: Fn({{ scalar_t }}) -> {{ scalar_t }},
+    {
+        Self::new(
+            {% for c in components %}
+                f(self.{{ c }}),
+            {%- endfor %}
+        )
+    }
+
     /// Creates a vector from the elements in `if_true` and `if_false`, selecting which to use
     /// for each element of `self`.
     ///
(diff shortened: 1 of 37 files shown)


Aceeri:
https://github.com/bevyengine/bevy/pull/14656/files#diff-0e640ab491c46a3af5c0e6c590467e40376348c3ccb7f4f69612765c96370fc4R381

Another example pointed out by @Jondolf

bitshifter (changes requested):
I was expecting someone to suggest this eventually :)

One change that I think would make sense is add it for all vector types, not just floating point vectors, I don't think there's any reason not to? This should just be a case of moving the function in codegen to a common area, it would also require tests (which again is mostly a case of moving them to a common location and casting the inputs to the scalar type in the test).

Aceeri:
> I was expecting someone to suggest this eventually :)
> 
> One change that I think would make sense is add it for all vector types, not just floating point vectors, I don't think there's any reason not to? This should just be a case of moving the function in codegen to a common area, it would also require tests (which again is mostly a case of moving them to a common location and casting the inputs to the scalar type in the test).

Oop ya that makes sense, I had thought this already did this but didn't check.

Aceeri:
Aaaa not sure what happened here

Aceeri:
Whoops, did a silly

bitshifter (approved):

bitshifter:
Just needs a rust fmt. It would be good if I could auto fmt PRs but I don't think that is possible

bitshifter:
Thanks for the PR!

PR #550, opened 2024-08-18
Reduce dot threshold at which lerp is used instead of slerp.

bitshifter:
Add lerp_impl which is reused to slerp and lerp but doesn't calc dot again.

### DIFF
--- a/codegen/templates/quat.rs.tera
+++ b/codegen/templates/quat.rs.tera
@@ -753,46 +753,7 @@ impl {{ self_t }} {
     #[inline(always)]
     #[must_use]
     fn lerp_impl(self, end: Self, s: {{ scalar_t }}) -> Self {
-        {% if is_scalar %}
-            let interpolated = self + ((end - self) * s);
-            interpolated.normalize()
-        {% elif is_sse2 %}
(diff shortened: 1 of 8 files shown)


codegen/templates/quat.rs.tera
```diff
@@ -897,57 +866,11 @@ impl {{ self_t }} {
...
-
-                let scale1 = i32x4_shuffle::<0, 0, 4, 4>(tmp, tmp);
-                let scale2 = i32x4_shuffle::<1, 1, 5, 5>(tmp, tmp);
-                let theta_sin = i32x4_shuffle::<2, 2, 6, 6>(tmp, tmp);
-
-                Self(f32x4_div(
-                    f32x4_add(f32x4_mul(self.0, scale1), f32x4_mul(end.0, scale2)),
-                    theta_sin,
-                ))
-            {% elif is_coresimd %}
```
bitshifter, line 922: Restore coresimd version and use coresimd vector sin

bitshifter: Left as is for now because sin is not part of core::simd, it's in std::simd.

PR #559, opened 2024-09-03
spirv: impl `Debug + Display`, fix `Affine::from_cols_array()`

Firestar99:
I would like to get a :+1: from the rust-gpu maintainers before this is merged, but feel free to already review this. Thanks :D

Added features for the spirv target:
* [rust-gpu v0.9.0](https://github.com/EmbarkStudios/rust-gpu/releases/tag/v0.9.0) added support for Debug and Display traits, this PR removes all the `#[cfg(not(target_arch = "spirv"))]` hiding the traits impls for the spirv target.
* Make `Affine::from_cols_array()` not use any slices internally and use arrays instead, as this caused compilation errors when these methods were used on the spirv target.

I tested this PR against embark's rust-gpu [master branch](https://github.com/EmbarkStudios/rust-gpu/commit/54f6978c25b7e168ded04e720b996625b3654ebd) and the [latest v0.9.0 release](https://github.com/EmbarkStudios/rust-gpu/releases/tag/v0.9.0) using their compiletest suite. 
(Note: test `member_ref_arg-broken` may fail on v0.9.0 due to a diff in stdout caused by newer spirv tools)

### DIFF
--- a/codegen/templates/affine.rs.tera
+++ b/codegen/templates/affine.rs.tera
@@ -733,7 +733,6 @@ impl PartialEq for {{ self_t }} {
     }
 }
 
-#[cfg(not(target_arch = "spirv"))]
 impl core::fmt::Debug for {{ self_t }} {
     fn fmt(&self, fmt: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         fmt.debug_struct(stringify!({{ self_t }}))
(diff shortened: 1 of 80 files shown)


bitshifter (approved):

bitshifter (approved):

eddyb:
> [rust-gpu v0.9.0](https://github.com/EmbarkStudios/rust-gpu/releases/tag/v0.9.0) added support for Debug and Display traits

To be clear, running the derived `impl`s isn't really supported, there's just some minimal `panic!` format string conversion to `debugPrintf`, mainly for simple scalars but I guess vectors are also supported?
(i.e. the only reason to have these `fmt::{Debug,Display}` `impl`s is so that e.g. `panic!("{some_vec4}")` type-checks, before it can get transformed into the equivalent `debugPrintf`)

eddyb (approved):

Firestar99:
@bitshifter could we get this merged? Thanks :D

bitshifter:
Sorry for the delay, GitHub didn't notify me that the PR had moved out of draft.

Issue #569, opened 2024-10-16
Improve documentation for projection matrix constructors


kovaxis:
Hello! I got very confused choosing the proper `Mat4::perspective_*` function. I think they could be documented a lot better. In particular, listing from what coordinate system they are mapping from, and into what coordinate system they are mapping into.

I've now done the research, and the transformations are:

```
perspective_rh_gl: X+ right, Y+ up, Z+ out of screen -> X+ right [-1, 1], Y+ up [-1, 1], Z+ into screen [-1, 1]
perspective_rh:    X+ right, Y+ up, Z+ out of screen -> X+ right [-1, 1], Y+ up [-1, 1], Z+ into screen [0, 1]
perspective_lh:    X+ right, Y+ up, Z+ into screen   -> X+ right [-1, 1], Y+ up [-1, 1], Z+ into screen [0, 1]
```

Where OpenGL uses `Z+ into screen [-1, 1]`, Metal/DirectX/WebGPU use `Z+ into screen [0, 1]` and Vulkan uses `Y+ down, Z+ into screen [0, 1]`.

The same issue applies to orthographic projections.

I propose the following to avoid breaking changes:
- Add `perspective_rh_vk` and `orthographic_rh_vk` functions, that map from `X+ right, Y+ up, Z+ out of screen` to `Y+ down, Z+ into screen [0, 1]` (what Vulkan expects).
- Document all of the projection functions with their corresponding modelview-coordinate-system -> graphics-api pairs.
  - `perspective_rh_gl`: `Z+ out of screen` -> GL
  - `perspective_rh_vk`: `Z+ out of screen` -> VK
  - `perspective_rh`: `Z+ out of screen` -> Metal/DX/WGPU
  - `perspective_lh`: `Z+ into screen` -> Metal/DX/WGPU
  - Corresponding `orthographic` functions.
- Further document the `perspective_infinite` and `perspective_infinite_reverse`.

There would be no `Z+ into screen` -> VK, just like currently there is no `Z+ into screen` -> GL. Maybe those two could be added? (`perspective_lh_gl`, `perspective_lh_vk`).

Actually, I collected my thoughts while writing this issue and I think I'll write up a PR.

bitshifter:
[...]

PR #570, opened 2024-10-16
Improve documentation for the `Mat4` projection functions

kovaxis:
Like I mentioned in #569, this PR adds more detailed documentation to the `Mat4::perspective_*` and `Mat4::orthographic_*` family of functions, as well as adding `Mat4::perspective_rh_vk` and `Mat4::orthographic_rh_vk`.

I initially wrote a very detailed doc explaining what world-view coordinates get mapped to what normalized-device-coordinates, but they look very out-of-place and are probably too much. Instead, I added a small subtext among the lines of `Useful to map the standard right-handed coordinate system into what OpenGL expects.`.

I haven't implemented the `*_rh_vk` functions or their tests yet. I would like to receive confirmation that this PR is ok before doing so.

### DIFF
--- a/codegen/templates/mat.rs.tera
+++ b/codegen/templates/mat.rs.tera
@@ -1716,7 +1716,22 @@ impl {{ self_t }} {
         Self::look_to_rh(eye, center.sub(eye), up)
     }
 
-    /// Creates a right-handed perspective projection matrix with [-1,1] depth range.
+    /// Creates a right-handed perspective projection matrix with `[-1,1]` depth range.
+    ///
+    /// # Interpretation
+    ///
+    /// This matrix an be interpreted as a projection that maps from the standard graphics "right-handed" coordinate system:
+    ///
+    /// - `X+` points right
+    /// - `Y+` points up
+    /// - `Z+` points *out of the screen*
+    ///
+    /// ...to the normalized device coordinate system that OpenGL expects:
+    ///
+    /// - `X+` points right, in the range `[-1, 1]`
+    /// - `Y+` points up, in the range `[-1, 1]`
+    /// - `Z+` points into the screen, in the range *`[-1, 1]`*
+    ///
     /// This is the same as the OpenGL `gluPerspective` function.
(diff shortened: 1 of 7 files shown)


bitshifter:
Hi, because there are 8 different projection methods currently adding a vk suffix will double that. At some point I'd like to revisit how this is done in glam, probably separating these out into separate modules for handedness and graphics api so users only get one in their project, although realistically I'm not sure when I will get around to that.

Happy to accept the  documentation improvements.

kovaxis:
> Hi, because there are 8 different projection methods currently adding a vk suffix will double that.

There would only be 2 new `_vk` functions (`perspective_rh_vk` and `orthographic_rh_vk`, not double the amount) just like there are only 2 `_gl` functions and not double the amount.

But sure, I've removed the functions. This is ready for review then.

> At some point I'd like to revisit how this is done in glam, probably separating these out into separate modules for handedness and graphics api so users only get one in their project, although realistically I'm not sure when I will get around to that.

`ultraviolet` has 4 modules (`rh_yup`, `lh_yup`, `rh_zup`, `lh_zup`), each with their own set of `(perspective,orthographic)_(wgpu_dx,gl,vk)` functions, if that's good reference. They also export by default the `rh_yup` to provide a good default for people who don't care (although to be honest I prefer blender-style `rh_zup`).

bitshifter:
> There would only be 2 new _vk functions (perspective_rh_vk and orthographic_rh_vk, not double the amount) just like there are only 2 _gl functions and not double the amount.

So far I've only provided the gl options as kind of compatibility with common gl functions that are in use, I kind of consider gl deprecated and therefore don't provide all of the possibly projection functions.

> ultraviolet has 4 modules (rh_yup, lh_yup, rh_ydown, lh_ydown), each with their own set of (perspective,orthographic)_(wgpu_dx,gl,vk) functions, if that's good reference. They also export by default the rh_yup to provide a good default for people who don't care (although to be honest I prefer blender-style rh_zup).

Yes I think ultraviolet might be the only library I've come across that actually tries to solve this. It might be a good approach for glam, I did want to put some thought into it myself but I've never managed to make the time.

Issue #562, opened 2024-09-13
Implement `{U,I}8Vec`


andrewgazelka:

bitshifter:
Do you have a use case for these or is this more because they aren't implemented yet?

andrewgazelka:
> Do you have a use case for these or is this more because they aren't implemented yet?

Use case. I want to use `U8Vec` because I want to represent chunk section coordinates in Minecraft which can be x,y,z in [0,16) so might as well use a `U8Vec` 🤷‍♂️

AlessandroCanossa:
Since I too need these vectors, I've been trying to implement them but I'm having some troubles with vec4 tests, such as

<details>

```
error: literal out of range for `i8`
    --> tests/vec4.rs:228:58
     |
228  |             assert_eq!($new(4 as $t, 16 as $t, 64 as $t, 256 as $t), a * a);
     |                                                          ^^^
...
2145 |     impl_vec4_signed_integer_tests!(i8, i8vec4, I8Vec4, I8Vec3, I8Vec2, BVec4, bvec4);
     |     --------------------------------------------------------------------------------- in this macro invocation
     |
```

```
error: literal out of range for `i8`
    --> tests/vec4.rs:448:45
     |
448  |             assert_eq!(a.element_product(), 210 as $t);
     |                                             ^^^
...
2145 |     impl_vec4_signed_integer_tests!(i8, i8vec4, I8Vec4, I8Vec3, I8Vec2, BVec4, bvec4);
     |     --------------------------------------------------------------------------------- in this macro invocation
     |
```

```
---- i8vec4::test_ops_propagated stdout ----
thread 'i8vec4::test_ops_propagated' panicked at glam-rs/src/i8/i8vec4.rs:868:23:
attempt to multiply with overflow
```
</details>

In your opinion, what would be the best way to solve these problems without risking breaking all the other tests? @bitshifter

bitshifter:
The macros are annoying to understand when they don't work.

The first one is from `test_ops`:

```
let a = $new(2 as $t, 4 as $t, 8 as $t, 16 as $t);
...
assert_eq!($new(4 as $t, 16 as $t, 64 as $t, 256 as $t), a * a);
```

The problem being that 16 * 16 is out of range. You could change this test to multiple by a different vector `b` so that the numbers don't get too large for `i8`,

The other failing test is `test_sum_product`

```
let a = $new(2 as $t, 3 as $t, 5 as $t, 7 as $t);
...
[...]

PR #575, opened 2024-10-28
Add i8,u8 support

AlessandroCanossa:
Solves #562


src/lib.rs
bitshifter, line 1: There's a doc comment at the top of this file that should have the new types added to it.

bitshifter:
Needs a `cargo fmt`, the codegen stuff is auto formatted but everything else isn't

bitshifter:
Thanks for all the work you've put into this!

PR #580, opened 2024-11-01
Add support for `rand`'s `Uniform` distribution to all Vec types

### LINKED ISSUE #579: Add support for `rand`'s uniform distribution, when using the `rand` feature
Hi! I'm practicing making games with `bevy`, and I'm making a simple snake game to do so. During this, I was searching for a way to point to a random tile within a grid.

I figured out that `glam` has the `rand` feature - but this only adds support for using `glam` types with the [`Standard` distribution](https://docs.rs/rand/0.8.5/rand/distributions/struct.Standard.html).

Since I had a limited grid size, I wanted to put bounds on the random numbers. This is possible using the `%` operator, but wouldn't necessarily result in a uniform distribution - some numbers may be ever so slightly more likely to appear than others.

I found that `rand` has a [distribution called `Uniform`](https://docs.rs/rand/0.8.5/rand/distributions/struct.Uniform.html), which is exactly what I want - I give it the bounds I want, and it generates a number between those two bounds, without the bias towards certain numbers.

Unfortunately, it seems `glam` doesn't implement the traits needed to use types like `UVec2` with the `Uniform` distribution. I'm making this request in hopes to rectify this issue.

----

**P.S.** Normally, I would be willing to make a PR for this sort of thing. However, I'm hesitant to say that I could deliver correct results for matrices or affine transformations; and for quaternions, I don't even know where to start. That said, if you're okay with a PR that only implements this for the Vec types, I can do that!

LikeLakers2:
Closes #579.

This PR makes the following changes to the `glam` crate:

* Adds support for `rand`'s `Uniform` distribution to all Vec types
* Reorganizes the imports for the `impl_rand.rs` file into their relevant macros, easing readability and maintainability

# This is still a Work-In-Progress

I have yet to do the following:

* Add tests for Float Vec types

I suspect that the code quality could also be improved in some places. If you spot any way to improve it, let me know! (or if you're bitshifter, feel free to commit directly to my PR!)

# Some notes

The `UniformVec*` structs used to handle uniform distribution are part of a non-public module, and thus cannot be navigated to in the documentation. While I can rectify this upon request, I don't believe it's necessary, since the way `rand` intends for the user to perform `Uniform` distribution is through the `rand::distributions::Uniform` type, rather than using the backend types directly.

I have not added (de)serialization support to the `UniformVec*` structs. However, `rand` offers (de)serialization of its distribution structs, so I am unsure if (de)serialization is wanted for `UniformVec*`.

### DIFF
--- a/src/features/impl_rand.rs
+++ b/src/features/impl_rand.rs
@@ -76,44 +76,95 @@ macro_rules! impl_int_types {
         impl UniformSampler for UniformVec2<$vec2, UniformInt<$t>> {
             type X = $vec2;
 
-            fn new<B1, B2>(_low: B1, _high: B2) -> Self
+            fn new<B1, B2>(low_b: B1, high_b: B2) -> Self
             where
                 B1: SampleBorrow<Self::X> + Sized,
                 B2: SampleBorrow<Self::X> + Sized,
             {
-                todo!()
+                let low = *low_b.borrow();
+                let high = *high_b.borrow();
+                assert!(low.x < high.x, "Uniform::new called with `low.x >= high.x");
+                assert!(low.y < high.y, "Uniform::new called with `low.y >= high.y");
+                Self {
+                    x_gen: UniformInt::new(low.x, high.x),
+                    y_gen: UniformInt::new(low.y, high.y),
+                    vec_type: PhantomData,
+                }
             }
 
-            fn new_inclusive<B1, B2>(_low: B1, _high: B2) -> Self
+            fn new_inclusive<B1, B2>(low_b: B1, high_b: B2) -> Self
             where
                 B1: SampleBorrow<Self::X> + Sized,
                 B2: SampleBorrow<Self::X> + Sized,
             {
-                todo!()
+                let low = *low_b.borrow();
+                let high = *high_b.borrow();
+                assert!(
+                    low.x < high.x,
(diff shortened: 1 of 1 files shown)


bitshifter:
This looks OK to me so far.

LikeLakers2:
@bitshifter Thank you. I'm working on cleaning up the code currently (as you can see in 9a683dc), and I still need to make tests for float types; but the main functionality is finished.

LikeLakers2:
Seeing as how I now have uniform distribution implemented (and basic tests for each) across all vec types except `Vec3A` (which doesn't have a uniform distribution implementation - not that it matters, since you can just convert a `Vec3`), I no longer consider this a draft.

Feel free to merge this at your leisure. If you have any suggestions or requests, I will try my best to implement them.

----

That said, while this implementation *is* working, I don't like the quality of the code I made. It doesn't help that I only just realized: I should have rewrote/refactored `impl_rand.rs`, as that would allow me to implement certain tests much more easily.

But if I were to do a refactor/rewrite of `impl_rand.rs`, I would prefer to do it in a separate PR, as I want to get these implementations out the door. The implementations as they stand are correct - or at least, as correct as `rand`'s own implementations.

bitshifter:
Thanks for this!

Issue #377, opened 2023-02-22
`slerp` for vectors (feature request)


kovaxis:
I'm in a situation where I would like to `slerp` two `Vec3`s, something like the following pseudocode:

```
fn slerp(from: Vec3, to: Vec3, alpha: f32):
    angle = angle_between(from, to) * alpha
    return from.rotate_by(amount: angle, plane: plane_between(from, to))
```

I think this can be implemented in terms of quaternions, but it is not directly obvious to me and I'm not sure if it is the most performant way either.

Does this fit in the scope of the library?

bitshifter:
You can create a Quat rotation between two vectors with https://docs.rs/glam/latest/glam/f32/struct.Quat.html#method.from_rotation_arc.

For slerp you need a start and end orientation to slerp between though, you can probably get that via:

```rust
let q0 = Quat::from_rotation_arc(FORWARD_DIR, from);
let q1 = Quat::from_rotation_arc(FORWARD_DIR, to);
let qa = q0.slerp(q1, alpha);
```

Where `FORWARD_DIR` is the bind pose of your object (i.e. what direction it faces with no rotation applied).

Whether that achieves what you want I am not sure. I would need to see some wider code context of how you are using this to determine if it makes sense to add a helper.

negamartin:
I was iterating on a dual contouring implementation, and I was interpolating normals between two points, so it really was just taking two unit `Vec3` and interpolating them across the unit sphere. I was doing a simple hack of `lerp`-ing them and then re-normalizing the result (`nlerp`), but I was getting odd results, so I wanted to quickly experiment by interpolating the normals "correctly".

I could kind of guess how to implement this in terms of quaternions, although it wasn't an instantly obvious solution as I'm not really familiar with quaternions. I did something along these lines:

```rust
Quat::IDENTITY.slerp(Quat::from_rotation_vec(from, to), alpha).mul_vec3(from)
```

I don't have the exact code I used because it turned out that normal interpolation was not the cause of my issues, and I quickly scrapped the code. I just thought that a proper `Vec3` slerp would be a nice quality-of-life addition to this library and it would have saved me some time. Also, at least according to [this stackoverflow answer](https://stackoverflow.com/a/67920029/5884836), a proper `Vec3::slerp` function can be more efficient than a quaternion-based implementation.

Of course, it's up to you whether it's worth it to include this functionality. As a precedent, Unity provides a `Vec3.slerp` function.

gkhk0:
[...]

PR #583, opened 2024-11-10
Add `Vec3::slerp`

tim-blackbird:
Based on [GLM](https://github.com/g-truc/glm/blob/33b4a621a697a305bc3a7610d290677b96beb181/glm/gtx/rotate_vector.inl#L5-L25)s implementation with added logic to handle:
- Non-normalized vectors
- Zero length vectors
- Parallel vectors, both in the same and opposite direction

This makes it identical in behavior to Godot's and Unity's slerp implementations

- Partially addresses #377

### Notes
- I wanted to add `Vec3::rotate_towards` as well making use of `slerp`, but I noticed that the behavior of `Vec2::rotate_towards` is different. `Vec2::rotate_towards` rotates `self` to be parallel to `rhs`, it doesn't ever become `rhs` If they have different lengths - I have updated the docs to reflect this -. A `slerp` based implementation *would* behave that way, and having different behavior between `Vec2` and `Vec3` for the same method would be undesirable.
- I benchmarked this against Godot's implementation using the infrastructure here and this performs a bit better on my machine (`33ns~` vs `36ns~`) while also accounting for parallel vectors in the opposite direction which Godot's implementation doesn't do.
- Is there desire for a more performant alternative method dedicated to normalized vectors?

### DIFF
--- a/benches/vec3.rs
+++ b/benches/vec3.rs
@@ -184,6 +184,15 @@ bench_select!(
     from => random_vec3
 );
 
+bench_trinop!(
+    vec3_slerp,
+    "vec3 slerp",
+    op => slerp,
+    from1 => random_vec3,
+    from2 => random_vec3,
+    from3 => random_f32
+);
+
 criterion_group!(
     benches,
     vec3_angle_between,
@@ -203,6 +212,7 @@ criterion_group!(
     vec3_to_array_into,
     vec3_to_rgb,
     vec3_to_tuple_into,
+    vec3_slerp,
 );
 
 criterion_main!(benches);
--- a/benches/vec3a.rs
+++ b/benches/vec3a.rs
@@ -154,6 +154,15 @@ bench_select!(
     from => random_vec3a
 );
 
+bench_trinop!(
+    vec3a_slerp,
+    "vec3a slerp",
+    op => slerp,
+    from1 => random_vec3a,
+    from2 => random_vec3a,
+    from3 => random_f32
+);
+
 criterion_group!(
     benches,
     vec3a_normalize_bench,
@@ -171,6 +180,7 @@ criterion_group!(
     vec3a_to_rgb,
     vec3a_to_tuple_into,
     vec3a_to_vec3,
+    vec3a_slerp,
 );
 
 criterion_main!(benches);
--- a/codegen/templates/vec.rs.tera
+++ b/codegen/templates/vec.rs.tera
@@ -2181,34 +2181,34 @@ impl {{ self_t }} {
         let self_length = self.length();
         let rhs_length = rhs.length();
         // Cosine of the angle between the vectors [-1, 1], or NaN if either vector has a zero length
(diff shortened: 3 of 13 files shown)


bitshifter:
Run out of time for tonight, but I did wonder what an example use case for slerping rotation and magnitude is? It's not something that occurred to me. Is that something Unity and Godot implementations do also?

There is some extra cost incurred if we can't assume that the input vectors are length zero and they behave differently to the Quat methods of the same name given we expect Quat to be unit length. Which isn't to say that scale shouldn't be handled, but possibly it is worth having versions that slerp or rotate with and without scale and the scale version is named differently from the Quat methods.

tim-blackbird:
> Ran out of time for tonight

Take your time! I'm not in a hurry - even if it may seem that way, haha

> Is that something Unity and Godot implementations do also?

Yes, both Unity and Godot interpolate the magnitude in their vector slerp implementations - which does make sense for a lerp in [spherical coordinates](https://en.wikipedia.org/wiki/Spherical_coordinate_system) -, *and* for the quat slerp implementation they require quaternions to be unit length, like ours.

Unity docs: [vector slerp](https://docs.unity3d.com/ScriptReference/Vector3.Slerp.html), [quat slerp](https://docs.unity3d.com/ScriptReference/Quaternion.Slerp.html)
Godot docs: [vector slerp](https://docs.godotengine.org/en/stable/classes/class_vector3.html#class-vector3-method-slerp), [quat slerp](https://docs.godotengine.org/en/stable/classes/class_quaternion.html#class-quaternion-method-slerp)

Unity's and Godot's implementations are likely the most well known implementations - *very* likely for Bevy's userbase -, and it still handles unit vectors as anyone would expect which makes me think calling this `slerp` is the right move.

This does run about `25%~` slower compared to the slerp method that assumes the inputs are unit vectors that I have locally. Which is a good reason to have a method - `slerp_unit` perhaps? - to reach for when the user knows they're only dealing with unit vectors in a hot loop

bitshifter:
I haven't had time to put much more thought into this, I did notice that Unreal doesn't have a method that slerps rotation and magnitude, it just has https://dev.epicgames.com/documentation/en-us/unreal-engine/API/Runtime/Core/Math/TVector/SlerpNormals and https://dev.epicgames.com/documentation/en-us/unreal-engine/API/Runtime/Core/Math/TVector/SlerpVectorToDirection (which maintains the magnitude of the input vector).

I would almost be inclined to only add `slerp_normals`, lerping magnitude in slerp seems like it would be fairly niche, but it could be added if someone wanted that specific functionality.

bitshifter:
I did think of one example where lerping the length might be useful, which is sometimes when dealing with velocity vectors it could be desirable to slerp towards a target velocity. Still, this feels a bit less common to me so I feel like offering a method that just slerps normals would be desirable as well, so people don't need to pay the cost of dealing with length if they don't need it.

bitshifter:
Decided to take this as is, `slerp_normals` could be added later.

Issue #516, opened 2024-05-26
Mutable Swizzles


the-ssd:
The main Idea is to be able to do this
```rust
let mut test_vec = Vec3::ZERO;
*test_vec.xy_mut() = Vec2::ONE;
assert_eq!(test_vec, Vec3::new(1.0, 1.0, 0.0))
```
Here is something that works, but only done for xy_mut()

```rust
trait MutSwizzleVec3<'a> {
    fn xy_mut(&'a mut self) -> Vec3XY<'a>;
}

impl<'a> MutSwizzleVec3<'a> for Vec3 {
    fn xy_mut(&'a mut self) -> Vec3XY<'a> {
        let xy = self.xy();
        Vec3XY {
            inner: self,
            thing: xy,
        }
    }
}

struct Vec3XY<'a> {
    inner: &'a mut Vec3,
    thing: Vec2,
}

impl<'a> core::ops::Deref for Vec3XY<'a> {
    type Target = Vec2;

    fn deref(&self) -> &Self::Target {
        &self.thing
    }
}

impl<'a> core::ops::DerefMut for Vec3XY<'a> {
    fn deref_mut(&mut self) -> &mut Vec2 {
        &mut self.thing
    }
}

impl Drop for Vec3XY<'_> {
    fn drop(&mut self) {
        self.inner.x = self.thing.x;
        self.inner.y = self.thing.y;
    }
}

#[test]
fn _test() {
    let mut test_vec = Vec3::ZERO;
    *test_vec.xy_mut() = Vec2::ONE;
    assert_eq!(test_vec, Vec3::new(1.0, 1.0, 0.0))
}
```

the-ssd:
Alternatively, it can be implemented like this. (But this doesn't allow passing `&mut Vec2` to a function, for example)
```rust
pub trait Vec3SwizzleSet {
    fn set_xz(&mut self, thing: Vec2);
    fn set_xy(&mut self, thing: Vec2);
    fn set_zy(&mut self, thing: Vec2);
    fn set_yz(&mut self, thing: Vec2);
}

impl Vec3SwizzleSet for Vec3 {
    fn set_xz(&mut self, thing: Vec2) {
        self.x = thing.x;
        self.z = thing.y;
    }

    fn set_xy(&mut self, thing: Vec2) {
        self.x = thing.x;
        self.y = thing.y;
    }

    fn set_zy(&mut self, thing: Vec2) {
        self.z = thing.x;
        self.y = thing.y;
[...]

PR #584, opened 2024-11-10
Add swizzle variants of `with_X` methods

tim-blackbird:
- Implement the idea from https://github.com/bitshifter/glam-rs/issues/516#issuecomment-2132421348

#### Example
```rs
Vec3::Y.with_xz(Vec2::ONE) // == Vec3::ONE
```

### Notes
- Do we want `set_X` swizzles also?
- >One complication is it would be good to support `Vec3` and `Vec3A` as the `thing` vector which will probably make the trait slightly more complex.
  
  Quote from the linked comment. Trivial to solve by taking `impl Into<{{ vec3_t }}>` though that would allow `[f32, f32, f32]`, etc. as well.


bitshifter:
Ah right, I guess I didn't see a need to port the swizzle test generation to tera when I migrated everything else. This is the old swizzlegen code before it was deleted https://github.com/bitshifter/glam-rs/blob/88e4538695cf1babddec1a9778510790cd1d7671/swizzlegen/src/main.rs

bitshifter:
Rather than adding it to tera I've restore enough of the old code to generate swizzle tests.

bitshifter:
https://github.com/bitshifter/glam-rs/tree/main/swizzlegen

tim-blackbird:
The tests use the values `11`, `12` and `13` for the right hand side `X`, `Y` and `Z` components respectively.
This hopefully makes verifying correctness easy enough :)
```rs
// Examples
assert_eq!(v.with_xyz(rhs3), vec4(11_f32, 12_f32, 13_f32, 4_f32));

assert_eq!(v.with_zx(rhs2), vec4(12_f32, 2_f32, 11_f32, 4_f32));
```

bitshifter (approved):

bitshifter:
I haven't reviewed every generated line but hopefully I've looked at enough to check things look correct AFAICT.

It took a little while to understand exactly what this does (having not used this in shaders) but it looks like it matches the https://registry.khronos.org/OpenGL/specs/gl/GLSLangSpec.4.60.pdf swizzle assignments.

The testing approach makes sense to me.

For set, it can be added later if there is a desire for it.

I think into `Into<{{ vec3_t }}>` would be more permissive than I would like. It could also be done by splitting the swizzle trait up https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fbc9b6031e9224e27c286d256101d412, but I'll leave that for now, it complicates things and the benefits would be very niche.

Thanks for this, it's a pretty fiddly thing to implement.

PR #586, opened 2024-11-11
Add `Vec3::rotate_towards`

tim-blackbird:
Added `Vec3::rotate_towards` with identical behavior to `Vec2::rotate_towards`

I added some extra tests to `Vec2::rotate_towards` as well, and adjusted the docs to be more accurate since the method does not interpolate the lengths of the vectors.

I noticed another issue with that. `Vec2::rotate_towards` returned `rhs` when the angle is too small. This means that only in that case the length of the resulting vector is equal to that of `rhs` instead of `self`.
It also differs from `Quat::rotate_towards` which returns `self` when the angle is too small, so I've updated it to return `self` instead.

### DIFF
--- a/benches/vec3.rs
+++ b/benches/vec3.rs
@@ -184,6 +184,15 @@ bench_select!(
     from => random_vec3
 );
 
+bench_trinop!(
+    vec3_rotate_towards,
+    "vec3 rotate_towards",
+    op => rotate_towards,
+    from1 => random_vec3,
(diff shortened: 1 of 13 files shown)


bitshifter:
I think perhaps returning the rhs when the angle is too small in both this and the quat version might be a better option.

bitshifter:
Replaced by #618 618

PR #589, opened 2024-11-21
Implement eigenvalue and eigenvector computation

### LINKED ISSUE #321: eigenvalues and eigenvectors
It would be nice to have methods to compute the eigenvalues and eigenvectors of matrices.

squaaawk:
A first implementation of eigenvalue computation in `glam`. Closes #321.

I intend to add eigenvector computation as well, but I would like some feedback on how to integrate this with `glam` first. This is a draft PR for two reasons:

1. The question of how to return complex values. Eigenvalues (and eigenvectors) can be complex, which `glam` does not support (and a full implementation of is beyond scope). Right now, I provide two methods for each matrix type: a general case eigenvalue function, which returns two-dimensional vectors storing the real and imaginary components, and a specialized eigenvalue function for symmetric matrices, where all eigenvalues are real. This storage of complex eigenvalues is not ideal, and would be more awkward for returning complex eigenvectors.

2. The question of where to integrate the code with `glam`'s types. The computation of eigenvectors, which has not been implemented yet, shares most of the work with the computation of eigenvalues. As such, the simple implementation of a `.eigenvalues()` and a `.eigenvectors()` method on each matrix type would be computationally inefficient. One option is to have one method which returns both, but eigenvalue computation can be more efficient if eigenvectors are not requested.

I have not looked too far into performance optimizations, just primarily correctness. There may be edge cases the current algorithm fails on I haven't encountered yet.

### DIFF
--- a/codegen/templates/mat.rs.tera
+++ b/codegen/templates/mat.rs.tera
@@ -1226,6 +1226,19 @@ impl {{ self_t }} {
         {% endif %}
     }
 
+    /// Returns the trace of `self`.
+    #[inline]
+    #[must_use]
+    pub fn trace(&self) -> {{ scalar_t }} {
+        {% if dim == 2 %}
+            self.x_axis.x + self.y_axis.y
+        {% elif dim == 3 %}
+            self.x_axis.x + self.y_axis.y + self.z_axis.z
+        {% elif dim == 4 %}
+            self.x_axis.x + self.y_axis.y + self.z_axis.z + self.w_axis.w
+        {% endif %}
+    }
+
     /// Returns the determinant of `self`.
     {%- if dim < 4 %}
     #[inline]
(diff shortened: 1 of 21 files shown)


bitshifter:
Hi @squaaawk , sorry it's taken me a long time to get to this. I think this raises a lot of questions that I hadn't really thought about, but it's made me think about the scope of what I want to support in glam and that this is probably out of scope. Looking at comparable game math libraries in Godot, Unity 3D or Unreal Engine, their math libraries don't support calculating eigenvectors and eigenvalues. As you point it also introduces a need for a complex number type, which glam doesn't currently have.

I think the best approach would be to create a separate crate which adds traits for calculating eigenvalues and eigenvectors, add num-complex as a dependency for a complex type implementation and implement these traits on glam types. 

I would be happy to add supporting methods to glam like the trace method added here.

PR #593, opened 2024-12-20
Implement manhattan and chebyshev distance for integer vectors

tguichaoua:
This implements #415 using the code generator.

### DIFF
--- a/codegen/README.md
+++ b/codegen/README.md
@@ -77,6 +77,7 @@ Each template starts with setting up a number of common variables based on the
 inputs from the `codegen` program. Commonly used variables are:
 
 * `self_t` - the name of the type being generated
+* `unsigned_scalar_t` - the unsigned version of `scalar_t` (e.g. `u8`, `u16`, `u32`)
 * `inner_t` - the inner storage type used by this type (e.g. `__m128` or
   `core::storage::XYZ<f32>`)
 * `deref_t` - the type used by the `Deref` and `DerefMut` implementation - not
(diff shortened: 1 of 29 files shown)


bitshifter (changes requested):
Thanks for the PR.

`abs_diff` on signed integers returns an unsigned integer so this will need a bit more work to code generate that .

The original #415 PR returned the result in a larger sized integer which for the Manhattan distance would help avoid overflow, however I guess for higher dimensions than 2 overflow would still be possible, so I'm not sure it's worth doing that. Also while glam doesn't currently have a `I128Vec*` or `U128Vec*` if they're ever added there is no `u256` primitive in Rust which would make returning a larger integer in with any 128 sizes integer vector not possible. Give that I'm leaning towards your implementation and not returning a larger integer type, but adding a comment to the method pointing out that the result could overflow so that the caller is aware. Do you have any thoughts on how overflow should be handled for these?

Could you also add some tests for these new methods as part of this PR.

tguichaoua:
> Do you have any thoughts on how overflow should be handled for these?

We should use unsigned integer as return type. As per the documentation of [`abs_diff`](https://doc.rust-lang.org/stable/std/primitive.i8.html#method.abs_diff):
> This function always returns the correct answer without overflow or panics by returning an unsigned integer.

For `manhattan_distance` I added `checked_manhattan_distance` which returns `None` if an overflow would occurs.

bitshifter (approved):

bitshifter:
Thanks!

PR #595, opened 2025-01-08
Add a compilation error if neither `std` nor `libm` features are specified

LikeLakers2:
Glam would previously cryptically error out during compilation, if neither the `std` nor `libm` features were specified. This is due to `no_std` being turned on, which removed many methods from the `f32` and `f64` primitives, while glam simultaneously would try to use those methods.

To solve this, I added a compilation error if no math backend is specified.

However, this `compile_error!()` does not silence the compilation errors generated by the use of non-existent methods. To silence them, I added a new module to `glam::f32::math` and `glam::f64::math` - one which defines the methods, but points them all to `unimplemented!()`.

The end result is that, if a backend is specified, glam compiles as normal - but if no backend is specified, glam returns a compilation error:
```shell
PS $root\bitshifter.glam-rs> cargo build --no-default-features
   Compiling glam v0.29.2 ($root\bitshifter.glam-rs)
error: You must specify a math backend using either the `std` feature or `libm` feature
   --> src\lib.rs:276:1
    |
276 | compile_error!("You must specify a math backend using either the `std` feature or `libm` feature");
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: could not compile `glam` (lib) due to 1 previous error

PS $root\bitshifter.glam-rs> cargo build --no-default-features --features std
   Compiling glam v0.29.2 ($root\bitshifter.glam-rs)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 8.90s

PS $root\bitshifter.glam-rs> cargo build --no-default-features --features libm
   Compiling glam v0.29.2 ($root\bitshifter.glam-rs)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 8.34s

PS $root\bitshifter.glam-rs> cargo build --no-default-features --features std,libm
   Compiling glam v0.29.2 ($root\bitshifter.glam-rs)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 8.31s
```

### DIFF
--- a/src/f32/math.rs
+++ b/src/f32/math.rs
@@ -140,7 +140,7 @@ mod libm_math {
     }
 }
 
-#[cfg(not(feature = "libm"))]
+#[cfg(all(not(feature = "libm"), feature = "std"))]
 mod std_math {
     #[inline(always)]
     pub(crate) fn abs(f: f32) -> f32 {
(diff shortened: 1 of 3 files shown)


bitshifter:
Makes sense, thanks!

LikeLakers2:
No problem! :D

PR #596, opened 2025-01-18
Add checked_* elementwise arithmetic operations for {I,U}Vec{2,3,4}

Dentosal:
For integer vectors, it's often useful to check if an operation would overflow. This helps with writing more robust code without having to do checks with `if` statements. Since we already have `wrapping_` and `saturating_` from Rust standard library, I figured having `checked_` would be welcome as well.

The test code is sadly very repetitive, as it's not generated from a template. Hopefully I got it all copy-pasted correctly!

### DIFF
--- a/tests/vec2.rs
+++ b/tests/vec2.rs
@@ -730,46 +730,46 @@ macro_rules! impl_vec2_signed_integer_tests {
         impl_vec2_signed_tests!($t, $new, $vec2, $vec3, $mask, $masknew);
 
         glam_test!(test_signum, {
-            assert_eq!($vec3::ZERO.signum(), $vec3::ZERO);
-            assert_eq!($vec3::ONE.signum(), $vec3::ONE);
-            assert_eq!((-$vec3::ONE).signum(), -$vec3::ONE);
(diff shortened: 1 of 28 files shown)


tests/vec2.rs
```diff
@@ -735,6 +735,43 @@ macro_rules! impl_vec2_signed_integer_tests {
             assert_eq!((-$vec3::ONE).signum(), -$vec3::ONE);
         });
 
+        glam_test!(test_checked_add, {
+            assert_eq!($vec3::MAX.checked_add($vec3::ONE), None);
```
bitshifter, line 739: these should be using $vec2 (the existing test_signum above also looks wrong)

tests/vec4.rs
```diff
@@ -2161,7 +2240,7 @@ mod i8vec4 {
         );
     });
 
-    glam_test!(test_wrapping_sub, {
+    glam_test!(test_wrapping_m, {
```
bitshifter, line 2243: This looks like an unintended change?

bitshifter (changes requested):

PR #597, opened 2025-01-20
Refactor `with_` and add new `set_` method for vector types

ganhaque:
## Objective:
The current behavior of the `with_` methods for vector types is using `mut self` type, which can be confusing since it relies on the `Copy` trait for implicit copy.

There are also no convenient `set_` method for method chaining pattern.

## Solution:
- Refactor the `with_` method of vector types to use `self` and create an explicit copy instead.
- Add a new `set_` method to allow for method chaining when mutating vector types.

### DIFF
--- a/codegen/templates/vec.rs.tera
+++ b/codegen/templates/vec.rs.tera
@@ -627,14 +627,16 @@ impl {{ self_t }} {
     /// Returns a new version of this {{ dim }}D vector with the given `{{ c }}` value.
     #[inline]
     #[must_use]
-    pub fn with_{{ c }}(&self, {{ c }}: {{ scalar_t }}) -> Self {
-        {% if self_t == "Vec3A" or self_t == "Vec4" %}
-            let mut new = *self;
-            new.{{ c }} = {{ c }};
-            new
-        {% else %}
-            Self { {{ c }}, ..*self }
-        {% endif %}
+    pub fn with_{{ c }}(self, {{ c }}: {{ scalar_t }}) -> Self {
+        Self::new(
+            {% for i in range(start = 0, end = dim) %}
+                {% if c == components[i] %}
+                    {{ c }},
+                {% else %}
+                    self.{{ components[i] }},
+                {% endif %}
+            {% endfor %}
+        )
     }
 {% endfor %}
 
--- a/codegen/templates/vec.rs.tera
+++ b/codegen/templates/vec.rs.tera
@@ -616,7 +616,7 @@ impl {{ self_t }} {
 
 
 {% for c in components %}
-    /// Sets the {{ c }} component of this {{ dim }}D vector
+    /// Sets the `{{ c }}` component of this {{ dim }}D vector
     #[inline]
     #[must_use]
     pub fn set_{{ c }}(mut self, {{ c }}: {{ scalar_t }}) -> Self {
@@ -624,7 +624,7 @@ impl {{ self_t }} {
         self
     }
 
-    /// Returns a new version of this {{ dim }}D vector with the given {{ c }} value.
+    /// Returns a new version of this {{ dim }}D vector with the given `{{ c }}` value.
     #[inline]
     #[must_use]
     pub fn with_{{ c }}(&self, {{ c }}: {{ scalar_t }}) -> Self {
--- a/codegen/templates/vec.rs.tera
+++ b/codegen/templates/vec.rs.tera
@@ -616,12 +616,25 @@ impl {{ self_t }} {
 
 
 {% for c in components %}
-    /// Creates a {{ dim }}D vector from `self` with the given value of `{{ c }}`.
+    /// Sets the {{ c }} component of this {{ dim }}D vector
     #[inline]
     #[must_use]
-    pub fn with_{{ c }}(mut self, {{ c }}: {{ scalar_t }}) -> Self {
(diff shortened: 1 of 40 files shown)


aloucks:
Usually I've seen and expect `with_` methods to take the object by value and return itself. Where as `set_` methods take a `&mut self` receiver. The convention in this proposed change does not make sense to me.  

These are all copy types anyway. The existing `with_` methods copy the item and return a modified copy. I don't see the point of adding the `set_` variants.

ganhaque:
I don't think the current `with_` method copy the item and return a modified copy?

Current:

```rust
/// Creates a 3D vector from `self` with the given value of `x`.
#[inline]
#[must_use]
pub fn with_x(mut self, x: f32) -> Self {
    self.x = x;
    self
}
```

New:
```rust
/// Returns a new version of this 3D vector with the given `x` value.
#[inline]
#[must_use]
pub fn with_x(&self, x: f32) -> Self {
    Self { x, ..*self }
}
```

I think both methods are pretty useful. The new `set_` (the old `with_` renamed) method provide a better way to initialize a vector type without the overhead of creating a new instance. The new `with_` method is for cloning without mutating the original.

```rust
// set_ method is better for initializing
let foo = Vec3::ZERO
    .set_x(1.0);

// with_ method is better for cloning
let bar = foo.with_y(1.0);
let x = foo.with_z(1.0);
```

`&mut self` for the `set_` method does make more sense than `mut self`. I didn't catch that when renaming the old `with_` method. I fixed it in the latest commit.

aloucks:
> I don't think the current `with_` method copy the item and return a modified copy?

Yes, it creates a copy. That's what `mut self` means as the receiver. It consumes the old one. If the type were not `Copy` the old one would no longer be available. A copy is NOT created when you use `&mut self` (a reference type).

```rust
#[test]
fn vec3_copy() {
    let a = Vec3::splat(1.0);
    let b = a.with_x(2.0);
    assert_eq!(1.0, a.x);
    assert_eq!(2.0, b.x);
}
```

ganhaque:
Ah, sorry for the misunderstanding, I though you meant create a copy as in creating a new instance with the same value.

>  A copy is NOT created when you use `&mut self` (a reference type).

Yeah, that is for the `set_` method which modify the original in place without consuming it. The new `with_` takes an immutable reference `&self` and returns a new instance with the modified value, while ensuring the original instance remains unchanged and available.

aloucks:
Your desired behavior for the way it works is how it **already** works. Look at the example unit test that I posted.

ganhaque:
Ah, I didn't understand how it work originally, thank you for explaining it. If both methods work, I think it would be better to **explicitly** create a copy instead of **implicitly** via the `Copy` trait? It would also be consistent with the swizzle `with_` methods.

aloucks:
I don't understand why you want to create `set_` methods for public fields and the `with_`  methods already work as expected. What problem are you trying to solve?

ganhaque:
The `set_` methods can be used for like

```rust
let mut foo = Vec3::ZERO;

// which allows for method chaining pattern
let modified_foo = foo
    .set_x(2.0)
    .set_y(3.0);
    
// instead of multiple statements
foo.x = 2.0;
foo.y = 3.0;
let modified_foo = foo;
```
It can also be used along with its swizzle variant (`set_xy`, etc.) if/when those are added. I just think it's a nice method to have for the method chaining coding style. I don't see the harm in adding it?

aloucks:
I'm trying to explain that it **already** works that way with the existing `with_` methods. When the receiver is `mut self`, it takes a **copy** of the receiver. The original item is **not** modified. This is idiomatic Rust code.

```rust
pub fn with_x(mut self, x: f32) -> Self {
    self.x = x;
    self
}
```

Using your example above:

`foo` does **not** need to be mutable because it's **not** being mutated.

```rust
#[test]
fn vec3_copy() {
    let foo = glam::Vec3::ZERO;

    let modified_foo = foo
        .with_x(2.0)
        .with_y(3.0);

    // foo has not changed
    assert_eq!(0.0, foo.x);
    assert_eq!(0.0, foo.y);
    assert_eq!(0.0, foo.z);

    // modified_foo has the new values specified by the `with_x` and `with_y` methods and the
    // original value for z
    assert_eq!(2.0, modified_foo.x);
    assert_eq!(3.0, modified_foo.y);
    assert_eq!(0.0, modified_foo.z);
}
```

ganhaque:
Sorry, I didn't update the original description. The `with_` method did work as intended. The change was just to make it create an explicit copy instead of implicitly via the `Copy` trait.

I was talking about the `set_` method instead, which is for mutating the vector. The `modified_foo` assignment was just to showcase mutating and assignment in a single statement.

```rust
// modify `foo` with new values
foo
    .set_x(2.0)
    .set_y(3.0);
   
// create a copy of `foo` with new values
foo
    .with_x(2.0)
    .with_y(3.0);
```

aloucks:
The fields are public. You can just do this:

```rust
foo.x = 2.0;
foo.y = 3.0;
```

ganhaque:
https://github.com/bitshifter/glam-rs/pull/597#issuecomment-2606217859

bitshifter:
glam does not provide `&mut self` methods as a convention and I don't see a compelling reason to add them here. The `with` changes seem unnecessary. It would be better to start a discussion about changes like this  to see if they are wanted.

Issue #598, opened 2025-01-27
Be more specific with angle parameter names


RobWalt:
Hey, this is just an idea. Feel free to discard it, if it's stupid.

Yesterday I was working on a little PR in bevy (https://github.com/bevyengine/bevy/pull/17545). I made a totally stupid rookie mistake and entered angles in degrees instead of radians in `Quat::from_axis_and_angle`.

It got me thinking: Would it be an option to rename the `angle` parameter to `radians`? I think it would make things more explicit and in some scenarios it could potentially be the one little extra hurdle that prevents a mistake like mine, since:

- the names are displayed in rendered doc comments in most editors
- the names are pre-filled when auto completing functions in some editors

Anyways, if you don't like it, feel free to close this issue without any second thoughts. I'll open a PR for this. Feel free to close it as well if you don't like the idea.

bitshifter:
Decided not to make this change

PR #599, opened 2025-01-27
docs: rename `angle` params to `radians`

RobWalt:
Implements #598


bitshifter:
Most of glam's code is generated, so to make this change you need to modify the templates and then run the code generator, see https://github.com/bitshifter/glam-rs/blob/main/codegen/README.md#running-codegen.

I guess this change seems reasonable, I prefer `angle` personally but `radians` is really synonymous with `angle` and would help with self documentation.

aloucks:
> @aloucks It's already documented in every function as `'angle' (in radians)`

It looks like it was missing in the docs for `Affine2::from_angle`, which happens to be the first item in the PR.

bitshifter:
Hmm, I think I'd be inclined to leave things as is in this case and fix up any documentation that's missing that angle is in radians.

PR #605, opened 2025-02-05
Update `rkyv` to latest 0.8 release

MarijnS95:
# Objective

Update `rkyv` to the `0.8` release, allowing downstream consumers to (de)serialize `glam` types with the new version.

## Code generation

N/A.

## Testing and linting

`cargo t -F rkyv` and `cargo t -F rkyv -F bytecheck`.

## Breaking changes

Users using the `rkyv` feature on `glam` **must** upgrade to `rkyv 0.8`.  There is no migration or fallback feature for `rkyv 0.7`.


src/features/impl_rkyv.rs
```diff
@@ -166,27 +167,36 @@ mod u64 {
...
     where
-        T: core::fmt::Debug + PartialEq + rkyv::Serialize<DefaultSerializer>,
-        T::Archived: core::fmt::Debug + PartialEq<T> + rkyv::Deserialize<T, DefaultDeserializer>,
+        T: core::fmt::Debug
+            + PartialEq
+            + rkyv::Portable
+            + for<'a> rkyv::Serialize<TestSerializer<'a>>,
+        T::Archived: core::fmt::Debug + PartialEq<T> + rkyv::Deserialize<T, TestDeserializer>,
+        // #[cfg(feature = "bytecheck")]
+        T::Archived: for<'a> rkyv::bytecheck::CheckBytes<TestValidator<'a>>,
```
MarijnS95, line 187: I don't think we can test `access()` and must always use `access_unchecked()`, because `cfg`'s don't work in `where` clauses and `access()` mandates `bytecheck` support.

src/features/impl_rkyv.rs
```diff
@@ -166,27 +167,36 @@ mod u64 {
...
     {
-        let mut serializer = DefaultSerializer::default();
-        serializer
-            .serialize_value(value)
-            .expect("failed to archive value");
-        let len = serializer.pos();
-        let buffer = serializer.into_serializer().into_inner();
-
-        let archived_value = unsafe { rkyv::archived_root::<T>(&buffer[0..len]) };
+        let buffer = rkyv::to_bytes(value).unwrap();
```
MarijnS95, line 189: The above was basically open-coding `rkyv::api::high::to_bytes()`, which we can easily inline.

MarijnS95:
@bitshifter the CI was enabling a feature on the `rkyv` crate that has been renamed.

On that note, should we tell docs.rs to enable various external features so that `impl Trait for SomeGlamType` is shown in our published docs?

Cargo.toml
```diff
@@ -41,21 +41,23 @@ fast-math = []
...
 libm = { version = "0.2", optional = true, default-features = false}
 
 [dev-dependencies]
 # rand_xoshiro is required for tests if rand is enabled
 rand_xoshiro = "0.7"
-# Set a size_xx feature so that this crate compiles properly with --all-targets --all-features
-rkyv = { version = "0.7", default-features = false, features = ["size_32"] }
+# Set a "pointer_width_xx "feature so that this crate compiles properly with --all-targets --all-features.
+# "alloc" is needed to support to_bytes()/deserialize() in tests.
+rkyv = { version = "0.8", default-features = false, features = ["pointer_width_32", "alloc"] }
```
bitshifter, line 60: This feature was enabled to work around a build error a while ago in https://github.com/bitshifter/glam-rs/commit/abf1a3ca762f5faa4c94998bcb10b0395962fd31, is it still necessary to do this with 0.8?

bitshifter:
Thanks. 

> On that note, should we tell docs.rs to enable various external features so that impl Trait for SomeGlamType is shown in our published docs?

I don't think I have an opinion one way or another, do you mean just for rkyv or for all dependencies?

I'll merge this but feel free to make a follow up PR.

MarijnS95:
@bitshifter correct, I expected that error to hold back merging but unfortunately it did not. 

This problem was already highlighted by me in: https://github.com/bitshifter/glam-rs/pull/605#discussion_r1942707852. The only other solution I have is dropping `bytecheck`-based tests.

PR #615, opened 2025-03-05
Add `min_position` and `max_position` for `Vec*`

EmbersArc:
# Objective

- Add `min_position` and `max_position` (also known as `argmin`/`argmax`) methods to vector types, as discussed in https://github.com/bitshifter/glam-rs/discussions/612. 

## Solution

- I was experimenting with something branchless (?) like `{{ self_t }}::splat(self.min_element()).cmpeq(self).bitmask().trailing_zeros() as usize` but then decided on a simpler solution.

## Code generation
 
- Updated the `vec.rs.tera`  template.

## Testing and linting

- Added tests for all vector sizes.

### DIFF
--- a/tests/vec2.rs
+++ b/tests/vec2.rs
@@ -510,6 +510,18 @@ macro_rules! impl_vec2_tests {
             assert_eq!($new(1 as $t, 2 as $t), b.max(a));
         });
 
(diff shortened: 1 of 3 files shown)


bitshifter:
Thanks!

PR #617, opened 2025-03-19
Add `nostd-libm` feature

bushrat011899:
# Objective

In [Bevy](https://github.com/bevyengine/bevy) we're working on adding `no_std` support, which has forced us to interact with the `std` or `libm` feature pair that Glam has. An issue we have is we want to support compiling `bevy` without default features, but our use of Glam makes that currently impossible. The workaround we've settled on is to just make all of our floating point maths related code optional, but that degrades the developer experience by making what was previously included functionality gated behind a seemingly unrelated question of `std` or `libm`.

Since `libm` compiles relatively instantly in the context of Bevy's other dependencies, and the fact that it's typically included by other dependencies already (e.g., `rand`), we'd like a way to enable `libm` purely for compatibility purposes, but still allow users to choose `std` as the floating point backend if they prefer it.

## Solution

- Add a new feature flag, `libm-fallback`, which includes `libm` as a dependency but only uses it if `std` is disabled.
- This feature is intended for libraries relying on Glam as a way to compile with default features disabled without denying the user the ability to choose their preferred backend.

## Code generation

- Modified code is not generated.

## Testing and linting

- CI

### DIFF
--- a/src/f32/math.rs
+++ b/src/f32/math.rs
@@ -32,7 +32,7 @@ fn acos_approx_f32(v: f32) -> f32 {
     }
 }
 
-#[cfg(feature = "libm")]
+#[cfg(any(feature = "libm", feature = "libm-fallback"))]
 mod libm_math {
     #[inline(always)]
     pub(crate) fn abs(f: f32) -> f32 {
@@ -236,7 +236,7 @@ mod std_math {
 
 // Used to reduce the number of compilation errors, in the event that no other
 // math backend is specified.
-#[cfg(all(not(feature = "libm"), not(feature = "std")))]
+#[cfg(all(not(feature = "libm"), not(feature = "std"), not(feature = "libm-fallback")))]
 mod no_backend_math {
     pub(crate) fn abs(_: f32) -> f32 {
         unimplemented!()
@@ -311,11 +311,11 @@ mod no_backend_math {
     }
 }
 
-#[cfg(feature = "libm")]
+#[cfg(any(feature = "libm", all(feature = "libm-fallback", not(feature = "std"))))]
 pub(crate) use libm_math::*;
 
 #[cfg(all(not(feature = "libm"), feature = "std"))]
 pub(crate) use std_math::*;
 
-#[cfg(all(not(feature = "libm"), not(feature = "std")))]
+#[cfg(all(not(feature = "libm"), not(feature = "std"), not(feature = "libm-fallback")))]
 pub(crate) use no_backend_math::*;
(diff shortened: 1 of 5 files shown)


bitshifter:
I'm a bit confused on how this works on the bevy side. Is it that you are adding the glam dependency with no default features but always use this libm-fallback feature flag in bevy?

src/f32/math.rs
```diff
@@ -32,7 +32,7 @@ fn acos_approx_f32(v: f32) -> f32 {
     }
 }
 
-#[cfg(feature = "libm")]
+#[cfg(any(feature = "libm", feature = "nostd-libm"))]
```
bitshifter, line 35: Should this also be `#[cfg(any(feature = "libm", all(feature = "nostd-libm", not(feature = "std"))))]` ? It will be unused if std is enabled.

src/f64/math.rs
```diff
@@ -1,4 +1,4 @@
-#[cfg(feature = "libm")]
+#[cfg(any(feature = "libm", feature = "nostd-libm"))]
```
bitshifter, line 1: Should be `#[cfg(any(feature = "libm", all(feature = "nostd-libm", not(feature = "std"))))]`?

src/lib.rs
```diff
@@ -274,7 +274,11 @@ The minimum supported Rust version is `1.68.2`.
     feature(portable_simd)
 )]
 
-#[cfg(all(not(feature = "std"), not(feature = "libm")))]
+#[cfg(all(
+    not(feature = "std"),
+    not(feature = "libm"),
+    not(feature = "nostd-libm")
+))]
 compile_error!("You must specify a math backend using either the `std` feature or `libm` feature");
```
bitshifter, line 282: This message needs updating to include `nostd-libm`

bitshifter (changes requested):
I left a few small questions and comments.

Also if you feel so inclined you could update the documentation on the available feature gates in the README.md and the module level documentation in src/lib.rs.

src/f32/math.rs
bushrat011899: That is correct, mistake on my part!

src/lib.rs
bushrat011899: Agreed. I've amended the wording to list all 3 features which can resolve this compiler error.

bushrat011899:
Done! Please let me know if there's anything else I can do to help!

bitshifter (approved):

bitshifter:
Thanks!

PR #630, opened 2025-04-24
Implement ser/de via `speedy` crate under feature gate

fu5ha:
# Objective

Have support for `speedy` with `glam`'s types

## Solution

I've moved the current implementation that exists in `speedy` under a `glam` feature flag over here to glam under a `speedy` feature flag instead. This is the better way for this to be set up in any case, and especially because `glam` releases significantly more often than `speedy` does.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,15 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog], and this project adheres to
 [Semantic Versioning].
 
+## Unreleased
+
+### Added
+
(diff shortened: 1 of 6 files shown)


bitshifter:
I haven't come across this crate before. Looking at the contributor list on it there's a lot of familiar faces though :)

fu5ha:
Any chance you could do a release with this soon? Would feel good to be able to switch to all released versions of glam and speedy for the first time in a while (and then release new versions of a couple dependent crates) 😄

bitshifter:
Yep, sure thing. I should be able to get that done sometime over the next couple of days.

Issue #659, opened 2025-07-18
`any_orthonormal_pair` and `cross` are not handedness-agnostic or documented


atlv24:
**Describe the bug**

Glam readme says it is handedness-agnostic, but `Vec3::any_orthonormal_pair` and `Vec3::cross` do not specify which handedness is produced and assume a handedness. (They are both right-handed)

**To Reproduce**

Read the docs on the methods

**Expected behavior**

Handedness is specified in either the docs or the function name as many functions in Mat3 do (`_rh` vs `_lh`)

**System information (please complete the following information):**

 - Any

**Additional context**

I imagine `cross_rh` `cross_lh` would be a little bit undesirable. But `any_orthonormal_pair_rh` vs `any_orthonormal_pair_lh` could be manageable. How do we communicate this?

atlv24:
I agree about the cross product, cross_lh and cross_rh would be confusing. I do still think it worthwhile to point out that the vector returned by cross follows the right-hand-rules, though.

Also, as I already read it and did the digging required to figure it out, some notes: the handedness is not documented anywhere in the JCGT/pixar paper or in Frisvad's original work, but it does talk about keeping the handedness consistent, fixing an issue with the original implementation where sometimes the opposite handedness is returned. I went to the original implementation source and saw that it was
1. generating any orthogonal vector
2. crossing it with the input vector for the third
and then I checked the cross product implementation, and it was the standard right-handed one. So it is a right-handed basis being constructed here.

bitshifter:
I found the original paper on Building an Orthonormal Basis from a Unit Vector by John F. Hughes and Tomas Moller, the abstract states: We show how to easily create a right-handed orthonormal basis, given a unit vector, in 2-, 3-, and 4-space. 

I have been pretty on the fence about adding _lh and _rh variants of this method, while you could argue that given there are lh and rh variants for look at and perspective functions I think for this low level basis creation function I don't want to do that. I have documented that the resulting orthonormal basis is right handed.

PR #660, opened 2025-07-19
Split any_orthonormal_pair into _rh and _lh variants

atlv24:
# Objective

- partial fix for #659

## Solution

- Split any_orthonormal_pair into _rh and _lh variants

## Code generation

- ya

## Testing and linting

- yup

## Breaking changes

- deprecations to avoid breaking


bitshifter:
Closing this as I've chosen to document the resulting basis rather than providing _lh and _rh methods in this instance.

PR #662, opened 2025-07-19
Specify that cross product is right handed in docs

atlv24:
# Objective

- Another part of #659 

## Solution

- Just document that its right-handed. I think cross product is probably used too often with the implicit right-handedness to bother specifying it as cross_rh. We should still point this out though.

## Code generation

- yar

## Testing and linting

- its docs

## Breaking changes

- its docs

### DIFF
--- a/src/f32/coresimd/vec3a.rs
+++ b/src/f32/coresimd/vec3a.rs
@@ -222,7 +222,7 @@ impl Vec3A {
         Self(dot3_into_f32x4(self.0, rhs.0))
     }
 
-    /// Computes the cross product of `self` and `rhs`.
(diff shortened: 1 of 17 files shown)


bitshifter:
For this one I don't think it's a change I want to merge. The glam docs document the API and in the case of some functions when the same inputs would give different outputs depending on the coordinate system. The docs generally avoid  explanations of linear algebra, as there are much better sources for that. To my knowledge, there is one common cross product equation, glam has not made a choice between left handed or right handed cross products that needs to be explained to users. So I think this falls on the side of documenting the math, rather than the API. It may be subjective but I also think stating that the cross product is right-handed could increase confusion rather than reducing it without a deeper explanation.

PR #668, opened 2025-09-11
impl `AsMut` for Quat

Firestar99:
# Objective

- `Quat` implements `AsRef` but not `AsMut`
- both `Vec` and `Mat` implement `AsRef` and `AsMut`
- (I don't *really* need it, it's just a detail I noticed while working on another PR)

## Solution

- adds `impl AsMut<[f32; 4]> for Quat`

## Code generation

was run

## Testing and linting

passes, though no new tests have been added. Vec and Mat also don't test this function.

### DIFF
--- a/src/f32/coresimd/quat.rs
+++ b/src/f32/coresimd/quat.rs
@@ -1233,6 +1233,14 @@ impl AsRef<[f32; 4]> for Quat {
     }
 }
 
+#[cfg(not(target_arch = "spirv"))]
+impl AsMut<[f32; 4]> for Quat {
+    #[inline]
+    fn as_mut(&mut self) -> &mut [f32; 4] {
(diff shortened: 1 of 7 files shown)


Firestar99:
I don't have a specific use-case, so feel free to close. And if anyone asks for this in the future, feel free to rebase and reopen :D

I just noticed it while searching for "spirv" feature gates, some PRs removing many of them coming soon! 

Anyway, if I were to submit some spirv specific PRs, how quickly would you be able to release a minor 0.30.6 patch? We're quite close to a rust-gpu release, and are currently debating whether to keep compat with older glam versions, requiring us to keep the `#[repr(SIMD)]` hacks, or remove them right now and require glam 0.30.6 or up.

PR #679, opened 2025-10-05
Add support for `zerocopy` using `zerocopy-derive`

lynlevenick:
# Objective

Adds support for the `zerocopy` crate, which fulfills a similar function to the `bytemuck` crate. `zerocopy` is slightly more widely used and I prefer their approach to memory safety.

Tries to select the correct subset of `zerocopy` derives based on which type is being generated.

This also tries to derive `ByteEq` and `ByteHash` when those are possible/sensible (that is, for non-floating point types that didn't already have an optimized implementation) though honestly I'm not sure if that's a breaking change and I don't need them myself so I would happily remove them if these are risky. If these are wanted, maybe `bytemuck`'s equivalents should also be set up if that's enabled?

### DIFF
--- a/src/bool/bvec2.rs
+++ b/src/bool/bvec2.rs
@@ -11,12 +11,7 @@ pub const fn bvec2(x: bool, y: bool) -> BVec2 {
 }
 
 /// A 2-dimensional `bool` vector mask.
-#[derive(Clone, Copy)]
-#[cfg_attr(
-    feature = "zerocopy",
-    derive(zerocopy::Immutable, zerocopy::KnownLayout)
-)]
-#[derive(PartialEq, Eq, Hash)]
+#[derive(Clone, Copy, PartialEq, Eq, Hash)]
 #[repr(C, align(1))]
 #[cfg_attr(target_arch = "spirv", rust_gpu::vector::v1)]
 pub struct BVec2 {
--- a/src/f32/coresimd/vec3a.rs
+++ b/src/f32/coresimd/vec3a.rs
@@ -26,12 +26,16 @@ pub const fn vec3a(x: f32, y: f32, z: f32) -> Vec3A {
 ///
 /// This type is 16 byte aligned.
 #[derive(Clone, Copy)]
+#[cfg_attr(feature = "bytemuck", derive(bytemuck::Pod, bytemuck::Zeroable))]
 #[cfg_attr(
     feature = "zerocopy",
-    derive(zerocopy::FromBytes, zerocopy::Immutable, zerocopy::KnownLayout)
+    derive(
+        zerocopy::FromBytes,
+        zerocopy::Immutable,
+        zerocopy::IntoBytes,
+        zerocopy::KnownLayout
+    )
 )]
-#[cfg_attr(feature = "bytemuck", derive(bytemuck::Pod, bytemuck::Zeroable))]
-#[cfg_attr(feature = "zerocopy", derive(zerocopy::IntoBytes))]
 #[repr(transparent)]
 pub struct Vec3A(pub(crate) f32x4);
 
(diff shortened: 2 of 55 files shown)


src/bool/bvec2.rs
```diff
@@ -11,7 +11,12 @@ pub const fn bvec2(x: bool, y: bool) -> BVec2 {
 }
 
 /// A 2-dimensional `bool` vector mask.
-#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+#[derive(Clone, Copy)]
```
bitshifter, line 14: It would be preferable for the generated code to keep `#[derive(Clone, Copy, PartialEq, Eq, Hash)]` in a single derive, even if it means some duplication in the template. My focus is on making the generated code more readable, as that's what most people will see.

src/f32/affine3a.rs
```diff
@@ -7,7 +7,24 @@ use core::ops::{Deref, DerefMut, Mul, MulAssign};
 ///
 /// This type is 16 byte aligned.
 #[derive(Copy, Clone)]
+#[cfg_attr(
+    feature = "zerocopy",
+    derive(zerocopy::FromBytes, zerocopy::Immutable, zerocopy::KnownLayout)
+)]
 #[cfg_attr(feature = "bytemuck", derive(bytemuck::AnyBitPattern))]
+#[cfg_attr(
```
bitshifter, line 15: Pretty minor but it would be good to keep the zerocopy attributes together (i.e. not have bytemuck in the middle),

templates/quat.rs.tera
```diff
@@ -124,7 +125,7 @@ pub struct {{ self_t }}(pub(crate) {{ simd_t }});
 
 impl {{ self_t }} {
     /// All zeros.
-    const ZERO: Self = Self::from_array([0.0; 4]);
+    pub const ZERO: Self = Self::from_array([0.0; 4]);
```
bitshifter, line 128: This seem unrelated to the goal of this PR? A zero quaternion is an invalid quaternion, which I think is why I didn't make this `pub`. I am using it internally so perhaps it's useful to others, but unless it's necessary for this PR I think this should be reverted and discussed separately.

src/f32/coresimd/vec3a.rs
```diff
@@ -26,7 +26,12 @@ pub const fn vec3a(x: f32, y: f32, z: f32) -> Vec3A {
 ///
 /// This type is 16 byte aligned.
 #[derive(Clone, Copy)]
+#[cfg_attr(
+    feature = "zerocopy",
+    derive(zerocopy::FromBytes, zerocopy::Immutable, zerocopy::KnownLayout)
+)]
 #[cfg_attr(feature = "bytemuck", derive(bytemuck::Pod, bytemuck::Zeroable))]
+#[cfg_attr(feature = "zerocopy", derive(zerocopy::IntoBytes))]
```
bitshifter, line 34: Where the condition is the same it would be preferable for these to be in a single derive. Because the generated code gets formatted, you might be able to put template conditions in the `derive` scope to conditionally include or exclude things like `IntoBytes` and it might still end up formatted nicely in the output.

src/i16/i16vec2.rs
```diff
@@ -14,8 +14,15 @@ pub const fn i16vec2(x: i16, y: i16) -> I16Vec2 {
 }
 
 /// A 2-dimensional vector.
-#[derive(Clone, Copy, PartialEq, Eq, Hash)]
+#[derive(Clone, Copy)]
+#[cfg_attr(
+    feature = "zerocopy",
+    derive(zerocopy::FromBytes, zerocopy::Immutable, zerocopy::KnownLayout)
+)]
+#[cfg_attr(not(feature = "zerocopy"), derive(PartialEq, Eq, Hash))]
```
bitshifter, line 22: Why are these excluded if using zerocopy? Do the zerocopy traits inherit from these?

bitshifter (changes requested):
Made some comments on trying to tidy up the generated code if you can.

PR #681, opened 2025-10-14
"Fix" zerocopy by not compiling it into the core-simd feature

lynlevenick:
# Objective

Fix support for the `zerocopy` crate.

Turns out that `zerocopy` doesn't have support for the `std::simd::Simd` type just yet? Frankly, either this or backing out the change entirely seem like appropriate solutions.

I'm so sorry for wasting your time with this extra work. I'm not sure zerocopy will support these types on a short timescale since their shape may still change, though I may spend time looking into that.

## Solution

Removes support for zerocopy under core-simd. Perhaps there should be a mutual exclusion of some features? Or just back out the previous PR entirely.

### DIFF
--- a/.tarpaulin.toml
+++ b/.tarpaulin.toml
@@ -33,7 +33,7 @@ exclude-files = [
 ]
 
 [core_simd]
-features = "core-simd approx bytemuck mint rand rkyv serde zerocopy debug-glam-assert"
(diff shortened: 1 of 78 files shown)


bitshifter:
Thanks for fixing that up!

PR #725, opened 2026-03-27
Fix typo in Mat4::transform_vector3

krakow10:
# Objective

- Fix typo

## Solution

- n


bitshifter:
Hey thanks for the PR. Because most glam code is generated by templates changes like this need to be made to the template and the files regenerated. For a small thing like this I've gone ahead and done this rather than asking you to go through that process.

PR #762, opened 2026-08-03
feat: make `Quat::to_array` a const fn

brianrodri:
# Objective

The `Vec` family provides symmetric `const from_array` and `const to_array` methods. `Quat` is inconsistent in this regard; only the `Quat::from_array` method is const. I want to fix this inconsistency so that I can convert to/from arrays at compile-time (which is helpful for writing const trait extensions on top of this library).

## Solution

Update the `Quat` template to make `to_array` const by borrowing `Vec`'s implementation:

https://github.com/bitshifter/glam-rs/blob/bd172a701971499191b8af85ed4d299e04057b08/templates/vec.rs.tera#L559-L569

## Code generation

- Most code in glam is generated by an offline tool. -- Ack'd
- Are you modify generated code? -- Yes
- Have you modified the templates and run the code generator? -- Yes

See https://github.com/bitshifter/glam-codegen/blob/main/README.md for details. -- Ack'd

## Testing and linting

- Your PR should include tests for any bug fixes or new features. -- N/A
- `cargo test` and `cargo clippy` must pass. -- Confirmed

You can run most of glam's tests locally by running `cargo run -p ci`. -- Done

### DIFF
--- a/src/f32/coresimd/quat.rs
+++ b/src/f32/coresimd/quat.rs
@@ -508,11 +508,11 @@ impl Quat {
         self.to_euler_angles(order)
     }
 
-    /// `[x, y, z, w]`
+    /// Converts `self` to `[x, y, z, w]`
     #[inline]
     #[must_use]
-    pub fn to_array(self) -> [f32; 4] {
-        [self.x, self.y, self.z, self.w]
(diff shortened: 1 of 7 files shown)


bitshifter:
Thanks!

Issue #770, opened 2026-08-17
Fix USizeVec2 layout assertion on 32-bit targets with cuda feature


heiserhorn:
On 32-bit targets such as i386 and armhf, compiling glam 0.30.1 with all features enabled fails in the compile-time layout test in src/usize.rs. USizeVec2 has repr(C) and repr(align(16)) when the cuda feature is enabled. Two usize fields occupy 8 bytes on 32-bit targets, but the 16-byte alignment makes the struct size 16 bytes, so the assertion size_of::<usize>() * 2 == size_of::<USizeVec2>() panics with Rust E0080. The assertion should account for the CUDA alignment on 32-bit targets. Debian is carrying a temporary packaging patch while this is resolved upstream. Reproducer: cargo test --all-features on a 32-bit target.

bitshifter:
Could you share the patch with me if it fixes the issue.

alexanderkjall:
Sorry for commenting on this after it's closed, but I did a testbuild from HEAD on a i386 vm and got these test errors:

<img width="1472" height="714" alt="Image" src="https://github.com/user-attachments/assets/65e0d051-fd2f-441e-b1aa-19030f28c100" />

With: cargo test --features cuda

heiserhorn:
Thanks for closing this issue. Debian's updated portability patch now covers the CUDA-enabled USizeVec2 layout assertion on 32-bit targets, and also fixes the I64Vec3/U64Vec3 alignment tests, which assumed 8-byte alignment on i386.

The Debian patch is available in MR !1326: https://salsa.debian.org/rust-team/debcargo-conf/-/merge_requests/1326

The patch has been refreshed with exact upstream context and forwarded here: https://github.com/bitshifter/glam-rs/pull/771

I verified rust-glam 0.30.1-4 with all features in an i386 sbuild: all Rust test groups passed, including the former alignment failures.

bitshifter:
Ah I didn't catch the u64 issue... I'll re-open. Also isize vectors have been added since 0.30 which have the same issues so trying to catch all of those as well.

heiserhorn:
Update: Debian unstable buildd now reports rust-glam 0.30.1-3 as Installed on both amd64 and i386, so the package builds successfully on those architectures and the earlier i386 alignment failures are resolved. The remaining upstream follow-up is PR #781, which forwards the platform-dependent I64Vec3/U64Vec3 alignment fix: https://github.com/bitshifter/glam-rs/pull/781

[...]

PR #771, opened 2026-08-18
Fix USizeVec2 layout checks on 32-bit targets

heiserhorn:
Fix the remaining 32-bit layout assertions for `USizeVec2` when the `cuda` feature is enabled.

On i386, the generated `tests/vec2.rs` integration test still compared `size_of::<usize>() * 2` (8 bytes) with the CUDA-aligned `USizeVec2` size (16 bytes), causing the all-features test to fail. The compile-time source assertion and integration test now account for the 32-bit CUDA layout while preserving the existing checks elsewhere.

This addresses issue #770. Debian observed the failure while building glam 0.30.1 on i386.

### DIFF
--- a/src/usize.rs
+++ b/src/usize.rs
@@ -10,8 +10,14 @@ pub use usizevec4::{usizevec4, USizeVec4};
 mod test {
     use super::*;
     mod const_test_usizevec2 {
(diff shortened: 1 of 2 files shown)


bitshifter:
Just needs `cargo fmt` to be run. You can run `cargo run -p ci` to run standard pre-commit checks.

PR #786, opened 2026-08-21
feat(vec3a): add explicit padding to scalar Vec3A for bytemuck Pod/Zeroable and zerocopy IntoBytes

### LINKED ISSUE #122: Bytemuck traits for SIMD types

bitshifter:
Fixes #122

## Problem

`Vec3A` is a 16-byte aligned type containing three `f32`s, so its scalar
(`scalar-math`, and the no-SIMD fallback) representation has 4 bytes of
trailing padding. That padding is uninitialized, so bytemuck could only derive
`AnyBitPattern` for scalar `Vec3A` (and `Mat3A`/`Affine3A`), not `Pod` +
`Zeroable`. The SIMD backends don't have this problem because their fourth
lane is a real value (`new` stores `[x, y, z, z]`).

## Change

Add an explicit `_w: f32` padding field to scalar `Vec3A`, initialized to `z`
to mirror the SIMD backends' `[x, y, z, z]` storage. This means every byte is
now initialized, which unlocks:

- `bytemuck::Pod` + `Zeroable` for scalar `Vec3A`, `Mat3A`, and `Affine3A`
  (`NoUninit` and `AnyBitPattern` come along automatically via bytemuck's
  blanket impls).
- `zerocopy::IntoBytes` for the same types (previously only
  `FromBytes`/`Immutable`/`KnownLayout`).

Scalar `Vec3A` now implements `PartialEq` manually so `_w` is ignored.

## Performance

The only added work in the materializing path is a single store that defines
the padding lane. When the result is inlined and only `x`/`y`/`z` are read,
that store is dead-code eliminated and the generated code is unchanged.

Probe:

```rust
#![allow(improper_ctypes_definitions)]
use glam::Vec3A;

#[no_mangle]
pub extern "C" fn probe_vec3a_add(a: Vec3A, b: Vec3A) -> Vec3A { a + b }

#[no_mangle]
pub extern "C" fn probe_vec3a_add_sum(a: Vec3A, b: Vec3A) -> f32 {
    let r = a + b;
    r.x + r.y + r.z
}
```

`probe_vec3a_add` before (`a867ca86`):

```asm
addss   xmm1, xmm3
addps   xmm0, xmm2
movss   dword ptr [rsp - 8], xmm1
movsd   xmm1, qword ptr [rsp - 8]   ; upper 4 bytes are uninitialized padding
ret
```

`probe_vec3a_add` after (`b465490b`):

```asm
addss   xmm3, xmm1
addps   xmm0, xmm2
movss   dword ptr [rsp - 8], xmm3
movss   dword ptr [rsp - 4], xmm3   ; <-- only new instruction: padding = z
movsd   xmm1, qword ptr [rsp - 8]
ret
```

`probe_vec3a_add_sum` is byte-for-byte identical before and after. The padding
store is eliminated because the value never escapes with its padding observed:

```asm
movaps  xmm4, xmm0
movlhps xmm4, xmm1
shufps  xmm4, xmm4, 232
movaps  xmm1, xmm2
movlhps xmm1, xmm3
psrlq   xmm0, 32
shufps  xmm1, xmm1, 232
addps   xmm1, xmm4
shufps  xmm2, xmm2, 85
addss   xmm0, xmm2
addss   xmm0, xmm1
shufps  xmm1, xmm1, 85
addss   xmm0, xmm1
ret
```

Reproduce with:

```sh
cargo rustc --release --example asm_probe --features scalar-math -- --emit asm -C llvm-args=--x86-asm-syntax=intel
```

### DIFF
--- a/src/f32/affine3a.rs
+++ b/src/f32/affine3a.rs
@@ -10,7 +10,7 @@ use zerocopy_derive::*;
 ///
 /// This type is 16 byte aligned.
 #[derive(Copy, Clone)]
-#[cfg_attr(feature = "bytemuck", derive(bytemuck::AnyBitPattern))]
+#[cfg_attr(feature = "bytemuck", derive(bytemuck::Pod, bytemuck::Zeroable))]
 #[cfg_attr(
     all(feature = "zerocopy", not(feature = "core-simd")),
     derive(FromBytes, Immutable, KnownLayout)
--- a/src/f32/affine3a.rs
+++ b/src/f32/affine3a.rs
@@ -13,19 +13,7 @@ use zerocopy_derive::*;
 #[cfg_attr(feature = "bytemuck", derive(bytemuck::Pod, bytemuck::Zeroable))]
 #[cfg_attr(
     all(feature = "zerocopy", not(feature = "core-simd")),
-    derive(FromBytes, Immutable, KnownLayout)
-)]
-#[cfg_attr(
-    all(
-        feature = "zerocopy",
-        any(
-            target_arch = "aarch64",
-            target_feature = "sse2",
-            target_feature = "simd128"
-        ),
-        not(any(feature = "core-simd", feature = "scalar-math"))
-    ),
-    derive(IntoBytes)
+    derive(FromBytes, Immutable, IntoBytes, KnownLayout)
 )]
 #[repr(C)]
 pub struct Affine3A {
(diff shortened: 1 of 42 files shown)


Firestar99:
@bitshifter this PR partially broke rust-gpu, due to `Vec3A` now looking like a `Vec4` with it containing 4 floats. The actual fallout is way smaller than I expected though, for now it just broke two intrinsics related to ray query (one of the two ray tracing extensions), and we I'm not even sure if anyone's using that. I won't ask for a revert just yet, this needs a bit more investigation first. 
See https://github.com/Rust-GPU/rust-gpu/issues/645

bitshifter:
@Firestar99 it's always possible to make some things conditional on `#[cfg(target_arch = "spirv")]` although having a divergence in bytemuck trait impls might be annoying if they're different between CPU and rust-gpu.

PR #789, opened 2026-08-23
fix(core-simd): make the core-simd backend build without std

### LINKED ISSUE #729: support `core-simd` in no-std
**What problem does this solve or what need does it fill?**

currently, the `core::simd` backend is only supported if `std` is enabled. this is because of `core::simd::StdFloat`, which assumes `std` is available. it would be interesting to use `core::simd` in no-std environments.

**Describe the solution you'd like**

allow usage of `core::simd` even without `std` by falling back to scalar `libm` code.

**Describe alternatives you've considered**

keeping it as is, or gating methods that depend on `StdFloat` behind `std` when `core-simd` is also enabled. 

**Additional context**

the use case is simply `no-std` with weird ISAs that don't have specific implementations in glam.

FadeHack:
# Objective

- Fixes #729.
- The `core-simd` backend doesn't compile in a `no_std` build. Enabling it without `std` currently fails with 18 errors.

To reproduce on nightly:

```
cargo build --no-default-features --features core-simd,libm,all-types
```

It all comes from one line. The vec template pulls in `use std::simd::StdFloat`, which only exists when `std` does, so you get the unresolved `std` import plus every `sqrt`, `round`, `floor`, `ceil`, `trunc` and `mul_add` call that trait was providing. Only `vec3a.rs` and `vec4.rs` are affected.

## Solution

Added a small `StdFloat` stand-in in `src/coresimd.rs` that's only compiled when `std` is off. It implements those six methods for `f32x4` lane by lane through the existing `math` module, so in a `no_std` build they go via `libm`. Only the methods glam actually calls are there, so it's a lot smaller than the real `StdFloat`.

When `std` is enabled nothing changes, the real `std::simd::StdFloat` is still re-exported and used, so there's no performance difference for existing builds.

The template import is now `crate::coresimd::StdFloat` instead of relying on the existing `coresimd::*` glob. An explicit import shadows the globs, so it can't go ambiguous if `core::simd` ever grows its own `StdFloat`.

## Code generation

Yes, `templates/vec.rs.tera` was modified and the generator was run. `src/coresimd.rs` is hand written. The generated diff is the one import line in `vec3a.rs` and `vec4.rs`.

## Testing and linting

The existing `core-simd` CI job only ever ran `cargo test --features core-simd`, which is the default `std` build, so it would never have caught this. I extended that command to also run the suite with `--no-default-features --features core-simd,libm,all-types`, so `cargo run -p ci -- core-simd` now covers both and this can't regress.

The full test suite passes on the `no_std` path, so the fallback matches what the `std` one produces. I also double checked the shim is really the code being used by temporarily sticking a `compile_error!` in it, which fires on the `no_std` build and not the `std` one.

- `cargo run -p ci` passes
- `cargo run -p ci -- core-simd` passes (both configurations)
- `cargo run -p ci -- msrv` passes
- scalar, neon, sse2, wasm simd128 and core-simd all still build


bitshifter:
While I'd be OK with committing this as is since it's an improvement on not compiling at all, currently this implementation looks like it will always calculate the `w` value, which for `Vec3A` is unused. Do you want to change this to only evaluate the used elements?

FadeHack:
Good catch, and it turned out to be worse than just `w`.

Rather than keep the `StdFloat` stand-in and special case the lane count inside it, I dropped it entirely. The template now picks between the SIMD call and the scalar one with a `cfg`, so without std these fall back to the same `math::` calls the scalar backend already uses, and only the elements that are actually used get evaluated.

The `w` one you spotted, but the bigger one was `length`, `length_recip` and `normalize`. Those do `dot3`/`dot4` into an `f32x4` and then sqrt the whole thing to pull out a single value, so lane by lane that was four sqrt calls to produce one number.

Per call, without std:

- `Vec3A` `round`/`floor`/`ceil`/`trunc`/`mul_add`: 4 -> 3
- `Vec3A` and `Vec4` `length`/`length_recip`/`normalize`: 4 sqrt -> 1
- `Vec4` `round`/`floor`/`ceil`/`trunc`/`mul_add`: still 4, all of them are used

Leaving `w` out looks safe on this backend. `BVec3A::bitmask` masks with `0x7` so everything mask based ignores it, `dot3` masks it out, and `Self::new` writes `w` itself so it never gets read. The scalar backend already does the same thing.

The std path is untouched, it still uses `std::simd::StdFloat` and the import is just cfg'd behind std now. I regenerated every file to check, and only the two coresimd vec files change.

PR #790, opened 2026-08-23
feat(mat): add row constructors and set_row to the matrix types

### LINKED ISSUE #639: Functions to work with rows
Matrix types (Mat3, DMat3, ..) seem to be missing functions to work with rows. This would be useful to have for various mathematical operations and also for interop with other libraries.

I would propose to add these functions:
* `Mat3::set_row(&mut self, i: usize, value: Vec3)`
* `Mat3::from_rows(row_1: Vec3, row_2: Vec3, row_3: Vec3) -> Self`
* `Mat3::from_rows_array(m: &[f32; 9) -> Self`
* `Mat3::from_rows_slice(slice: &[f32]) -> Self`
* `Mat3::to_rows_array(&self) -> [f32; 9]`

Similarly for all of these matrix types:
* Mat2
* DMat2
* Mat3
* DMat3
* Mat4
* DMat4

Happy to create a PR if this is something which you would appreciate.

FadeHack:
# Objective

- Fixes #639.
- The matrix types have `row()` for reading, but no way to build a matrix from rows or write one back. Row major is what you usually get out of file formats and other math libraries, so everyone ends up writing the same transpose glue themselves.

Adds the functions from the issue to `Mat2`, `Mat3`, `Mat3A`, `Mat4` and the `DMat` equivalents:

- `from_rows`
- `from_rows_array`
- `from_rows_slice`
- `to_rows_array`
- `set_row`

## Solution

You mentioned in the issue you were fine with these as long as they carry a doc comment reminding people glam stores columns and that rows cost more, so each one says that and links to the column equivalent to use instead when you don't need rows.

On the overhead itself, it turned out to be cheaper than expected. The array and slice constructors are just an index permutation into the existing `Self::new`, and `to_rows_array` permutes `to_cols_array`, so none of them actually call `transpose` and all four stay `const` like their column counterparts. `set_row` writes straight into the columns, so it's the only one that can't be const.

I left out `from_rows_array_2d` and `to_rows_array_2d` since they weren't in the issue, happy to add them if you want the mirror to be complete.

## Code generation

Yes. Everything is in `templates/mat.rs.tera` and the generator was run, which is where the 19 generated matrix files in the diff come from.

## Testing and linting

Added a `test_matN_rows` test to `tests/mat2.rs`, `tests/mat3.rs` and `tests/mat4.rs`, so it runs for every type variant including `Mat3A` and the `DMat` ones. They cover:

- the row major array round trip through `from_rows_array` / `to_rows_array` / `from_rows_slice`
- that `to_rows_array()` agrees with `transpose().to_cols_array()`
- `from_rows` rebuilding a matrix from its own `row()` values
- `set_row` round tripping with `row()`
- the const paths, so constness is actually exercised and not just declared
- out of bounds panics for `set_row` and a too short slice

- `cargo run -p ci` passes
- `cargo run -p ci -- msrv` passes, so the new const fns are fine on 1.68.2
- checked scalar, neon, sse2, wasm simd128 and core-simd all build, including the `Mat2` SIMD case that goes through `Deref` rather than real fields

### DIFF
--- a/templates/mat.rs.tera
+++ b/templates/mat.rs.tera
@@ -297,7 +297,7 @@ impl {{ self_t }} {
 
     /// Creates a {{ nxn }} matrix from {{ dimension_in_full }} row vectors.
     ///
-    /// Note that `glam` stores matrices in column major order, so the data given here is
+    /// Note matrices are stored in column major order, so the data given here is
     /// transposed on the way in. Prefer [`Self::from_cols`] when the data is already
     /// laid out as columns.
     #[inline(always)]
@@ -395,7 +395,7 @@ impl {{ self_t }} {
 
     /// Creates a {{ nxn }} matrix from a `[{{ scalar_t }}; {{ size }}]` array stored in row major order.
     ///
-    /// Note that `glam` stores matrices in column major order, so the data given here is
+    /// Note matrices are stored in column major order, so the data given here is
     /// transposed on the way in. Prefer [`Self::from_cols_array`] when the data is already
     /// in column major order.
     #[inline]
@@ -412,7 +412,7 @@ impl {{ self_t }} {
 
     /// Creates a `[{{ scalar_t }}; {{ size }}]` array storing data in row major order.
     ///
-    /// Note that `glam` stores matrices in column major order, so this transposes the data
+    /// Note matrices are stored in column major order, so this transposes the data
     /// on the way out. Prefer [`Self::to_cols_array`] when column major data will do.
     #[inline]
     #[must_use]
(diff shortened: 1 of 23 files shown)


templates/mat.rs.tera
```diff
@@ -295,6 +295,30 @@ impl {{ self_t }} {
         {% endif %}
     }
 
+    /// Creates a {{ nxn }} matrix from {{ dimension_in_full }} row vectors.
+    ///
+    /// Note that `glam` stores matrices in column major order, so the data given here is
```
bitshifter: A small request - maybe rephrase this as `Note matrices are stored in column major order` because some crates re-export glam types (e.g. bevy) and some users of those crates may not be aware that they're actually using glam.

FadeHack: Makes sense, done in 9b22213. They all read "Note matrices are stored in column major order" now, so it doesn't matter which crate you got the types from.

bitshifter:
For this one I'm having second thoughts on if I actually want to do this. I'll comment on the issue to discuss it when I get around to it.

PR #792, opened 2026-08-25
Bench improvements

bitshifter:
Switch from callgrind to cachegrind for gungraun benches and wrap code to be benched with `client_requests::cachegrind::start_instrumentation()` and `client_requests::cachegrind::stop_instrumentation()` which is hopefully slightly more accurate.

Also add option the the ci bench command to bench a single backend.


bitshifter:
Cachegrind benches ended up having more overhead due to capturing the stop client-request block. It's also a bit more complex dependency wise, so sticking with callgrind.

PR #795, opened 2026-08-25
feat: add float_eq support behind an optional feature

### LINKED ISSUE #470: Feature request: Support for the `float_eq` crate
[`float_eq`](https://docs.rs/float_eq/latest/float_eq/) is an alternative to `approx`, but it has some features that `approx` lacks, most importantly being [derive macros](https://jtempest.github.io/float_eq-rs/book/how_to/derive_the_traits.html). It would be greatly appreciated if you could add support so that users of `glam` can also derive floating-point equality for their types.

ChrisJr404:
# Objective

- Add optional `float_eq` support for the glam types, matching the existing `approx` feature.
- Closes #470.

Each type forwards to its component array (`to_array` / `to_cols_array`) and defers to the `[T; N]` impls in `float_eq`, so the per-field tolerances and debug output line up with the components. I wired it into the ci feature lists next to approx and added a test covering the abs, relative and ulps comparisons for every type.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -83,6 +83,7 @@ encase = { version = "0.12", optional = true, default-features = false }
 zerocopy = { version = "0.8", optional = true, default-features = false, features = ["simd"] }
 zerocopy-derive = { version = "0.8", optional = true, default-features = false }
 arbitrary = { version = "1.4.2", default-features = false, optional = true }
+float_eq = { version = "1.0.1", default-features = false, optional = true }
 
 [dev-dependencies]
 # rand_xoshiro is required for tests if rand is enabled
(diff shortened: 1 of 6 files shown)


bitshifter:
Are you wanting to use `float_eq` with glam? I ask because I've left #470 open for a while to see if there's more interest. At this point the approx crate is a lot more popular so I'm wondering how much genuine demand there is for float_eq. I would prefer not to add a new dependency if there's not much demand for it, but aside from downloads on crates.io it's hard to tell.

bitshifter:
If you rebase from main hopefully those bench actions should pass.

ChrisJr404:
Rebased on main, so the bench actions should pass now.

PR #796, opened 2026-08-25
Add float move_towards methods

CodedNil:
# Objective

Glam implements move_towards for Vec2 Vec3 Vec4 but not for Floats

## Solution

Added very simple implementation of move_towards for Float, with tests

### DIFF
--- a/src/f32/float.rs
+++ b/src/f32/float.rs
@@ -37,4 +37,14 @@ impl FloatExt for f32 {
     fn saturate(self) -> Self {
         self.clamp(0.0, 1.0)
     }
+
+    #[inline]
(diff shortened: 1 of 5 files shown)


bitshifter:
I'm curious, what do you want to use this for?

CodedNil:
I use it extensively in my ui app currently, similar in fashion to how I'd use lerp except for linear stepped changes. For animating values towards a target.
```rust
self.icon_morph = self.icon_morph.move_towards(1.0, speed * 1.5);

self.action_hover = self.action_hover.move_towards(f32::from(response.hovered()), context.frame.delta_time / 0.12);
```

PR #797, opened 2026-08-25
feat: add smoothstep methods

CodedNil:
# Objective

Smoothstep is a common standardised function available in shaders, but not currently available in glam unlike map_toward lerp inverse_lerp etc.

## Solution

Adds smoothstep methods to floats and vecs, it uses (x, edge0, edge1) signature which differs from opengl but fits with the rust standards better I think. as you might do x.smoothstep(edge0, edge1) and it doesn't really read correctly to do edge0.smoothstep(edge1, x)

### DIFF
--- a/templates/vec.rs.tera
+++ b/templates/vec.rs.tera
@@ -2114,6 +2114,19 @@ impl {{ self_t }} {
         Self::select(rhs.cmplt(self), Self::ZERO, Self::ONE)
     }
 
+    /// Performs Hermite interpolation between `0.0` and `1.0` using `x` normalized to `[edge0, edge1]`.
+    ///
+    /// This is equivalent to `t * t * (3.0 - 2.0 * t)`, where `t` is clamped to `[0.0, 1.0]`.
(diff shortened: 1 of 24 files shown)


templates/vec.rs.tera
```diff
@@ -2114,6 +2114,18 @@ impl {{ self_t }} {
         Self::select(rhs.cmplt(self), Self::ZERO, Self::ONE)
     }
 
+    /// Performs Hermite interpolation between `0.0` and `1.0` using `x` normalized to `[edge0, edge1]`.
+    ///
+    /// This is equivalent to `t * t * (3.0 - 2.0 * t)`, where `t` is clamped to `[0.0, 1.0]`.
+    /// Results are undefined if any element of `edge0` is greater than or equal to the corresponding
```
bitshifter, line 2120: Do you think it would be worth adding a `glam_assert` to check this?

CodedNil: I have pushed a change with that, hopefully done in the way you were expecting? Added a few more tests throughout.

bitshifter:
Thanks!

PR #824, opened 2026-09-04
perf(quat): avoid acos in is_near_identity

pnn64:
Glam has been a big help while optimizing [DeadSync](https://github.com/pnn64/deadsync), my Rust rewrite of StepMania/ITGmania, so I went looking for small wins I could potentially contribute upstream.

Is there a reason why we don't compare `abs(w)` directly here instead?

Feel free to close this if it has been tried previously or if there is a reason to retain the angle calculation.

`Quat::is_near_identity` currently calculates:

```rust
2.0 * acos(abs(w)) < threshold_angle
```

Because `acos` is monotonically decreasing, this can instead be expressed as a direct comparison against the corresponding cosine threshold.

## Solution

Replace the approximate `acos` calculation with:

```rust
abs(w) > threshold
```

The existing f32 and f64 tolerances are retained. Boundary tests cover:

- The exact threshold
- The next representable value inside the threshold
- Both quaternion signs
- NaN inputs

## Benchmarks

On my machine:

| Metric | Before | After | Change |
|---|---:|---:|---:|
| Criterion time | 7.8916 ns | 2.4259 ns | -69.84% |
| Callgrind instructions | 38 | 7 | -81.6% |
| DHAT allocations | 0 | 0 | No change |

## Code generation

The quaternion template was modified and the generated f32 SIMD/scalar and f64 implementations were regenerated.

## Testing and linting

- Quaternion tests: 83 passed
- All feature-test configurations passed
- Documentation and no-std checks passed
- Rust 1.68.2 MSRV checks passed
- Patch-scoped clippy passed with warnings denied
- Quaternion code-generation check passed

🤖 AI disclosure: I used Codex with GPT-5.6 Sol at xhigh reasoning effort to help identify the optimization and construct the benchmark. I have reviewed and understand the resulting implementation.

### DIFF
--- a/benches/gungraun.rs
+++ b/benches/gungraun.rs
@@ -416,6 +416,12 @@ fn quat_inverse(q: Quat) -> Quat {
 }
 
 #[library_benchmark]
+#[bench::args(quat())]
+fn quat_is_near_identity(q: Quat) -> bool {
+    black_box(q.is_near_identity())
+}
+
+#[library_benchmark]
 #[bench::positive_dot(quat(), quat_rot_x_180(), bb_f32())]
 #[bench::negative_dot(quat(), quat_rot_x_240(), bb_f32())]
 fn quat_lerp(q1: Quat, q2: Quat, t: f32) -> Quat {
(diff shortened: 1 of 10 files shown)


bitshifter:
Hah, you make a good point. Thanks!

PR #827, opened 2026-09-05
perf(mat4): specialize affine determinant in SRT decomposition

pnn64:
Following #824, could we also skip the full 4x4 determinant when decomposing an affine matrix?

## Solution

This uses the 3x3 linear part instead, keeping the existing expansion order to avoid changing behavior near underflow. Templates, generated code, tests, and benchmarks are included.

## Benchmarks

On my Xeon E5-2696 v4:

| Measurement | Before | After | Change |
|---|---:|---:|---:|
| 256-matrix batch | 4.879 µs | 3.705 µs | -24.1% |
| Throughput | 52.47 M/s | 69.10 M/s | +31.7% |
| Cycles/matrix | 62.608 | 49.313 | -21.2% |
| Instructions/matrix | 148.200 | 119.634 | -19.3% |
| Allocations | 0 | 0 | Unchanged |

Timing: Criterion on Windows. Counters: five alternating, core-pinned runs under WSL.

## Testing and linting

- Three million bit-for-bit comparisons passed.
- All nine feature-test configurations passed.
- Glam-specific Clippy, formatting, and generated-matrix checks passed.
- Existing Windows workspace-check issues remain.

🤖 AI disclosure: I used Codex with GPT-6 Astra at xhigh reasoning effort to help identify the optimization and construct the benchmark. I have reviewed and understand the resulting implementation.

### DIFF
--- a/templates/mat.rs.tera
+++ b/templates/mat.rs.tera
@@ -725,10 +725,12 @@ impl {{ self_t }} {
     #[inline]
     #[must_use]
     pub fn to_scale_rotation_translation(&self) -> ({{ vec3_t }}, {{ quat_t }}, {{ vec3_t }}) {
+        // The full matrix can be singular even when its linear part is not.
+        glam_assert!(self.determinant() != 0.0);
+
         // The determinant of an affine matrix is the determinant of its linear part.
         // Expand along the first column, like `determinant()`, to preserve underflow behavior.
         let det = self.x_axis.xyz().dot(self.y_axis.xyz().cross(self.z_axis.xyz()));
-        glam_assert!(det != 0.0);
 
         let scale = {{ vec3_t }}::new(
             self.x_axis.length() * math::signum(det),
--- a/templates/mat.rs.tera
+++ b/templates/mat.rs.tera
@@ -725,7 +725,9 @@ impl {{ self_t }} {
     #[inline]
     #[must_use]
     pub fn to_scale_rotation_translation(&self) -> ({{ vec3_t }}, {{ quat_t }}, {{ vec3_t }}) {
-        let det = self.determinant();
+        // The determinant of an affine matrix is the determinant of its linear part.
+        // Expand along the first column, like `determinant()`, to preserve underflow behavior.
+        let det = self.x_axis.xyz().dot(self.y_axis.xyz().cross(self.z_axis.xyz()));
         glam_assert!(det != 0.0);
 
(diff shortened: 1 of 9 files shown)


templates/mat.rs.tera
```diff
@@ -725,7 +725,9 @@ impl {{ self_t }} {
     #[inline]
     #[must_use]
     pub fn to_scale_rotation_translation(&self) -> ({{ vec3_t }}, {{ quat_t }}, {{ vec3_t }}) {
-        let det = self.determinant();
+        // The determinant of an affine matrix is the determinant of its linear part.
+        // Expand along the first column, like `determinant()`, to preserve underflow behavior.
+        let det = self.x_axis.xyz().dot(self.y_axis.xyz().cross(self.z_axis.xyz()));
         glam_assert!(det != 0.0);
```
bitshifter: The purpose of the `glam_assert` is to panic if `self` is not an affine matrix, I assume this change would mean this check would no longer be able to determine that 100% of the time?

pnn64: Good catch, thanks! I've restored the full determinant check inside `glam_assert!` and added a regression case for a zero fourth column. The optimized path is unchanged when assertions are disabled.

templates/mat.rs.tera
```diff
@@ -725,8 +725,12 @@ impl {{ self_t }} {
     #[inline]
     #[must_use]
     pub fn to_scale_rotation_translation(&self) -> ({{ vec3_t }}, {{ quat_t }}, {{ vec3_t }}) {
-        let det = self.determinant();
-        glam_assert!(det != 0.0);
+        // The full matrix can be singular even when its linear part is not.
+        glam_assert!(self.determinant() != 0.0);
```
bitshifter: Thinking about it, I think calculating the 4x4 determinant doesn't really add much so it's probably best to go back to how you originally wrote it. In some other places I check that the last row is Vec4::W:

```
glam_assert!(self.row(3).abs_diff_eq({{ vec4_t }}::W, 1e-6));
```

I think the combination of the 3x3 determinant plus this check would be better than the 4x4 determinant check that I was doing (as separate checks).

templates/mat.rs.tera
bitshifter: I think I might actually go back to how you originally had it and check the self.row(3) == Vec4::W instead which is probably a more targeted affine check that doing a 4x4 det, but I can do that post merge.

Issue #830, opened 2026-09-06
Consider removing simd backed Mat2


bitshifter:
There are not significant gains from having a simd backed Mat2. A lot of the gungraun benches are within +-2 instructions of noise. Being simd backed as the downside of being 16 byte aligned, so types like Affine2 are 25% padding due to Mat2 alignment.

See https://github.com/bitshifter/glam-bench-baselines/blob/main/x86_64.md and https://github.com/bitshifter/glam-bench-baselines/blob/main/aarch64.md, neon in particular has a consistently higher instruction count than the scalar implementation.

Need to run the criterion benchmarks as well to check if they behave similarly to the gungraun benchmarks.

This would be a breaking change due to alignment and affine2 padding changes.

PR #835, opened 2026-09-07
perf(mat3a): avoid intermediate Mat2 in transform_vector2

pnn64:
`Mat3A::transform_vector2` can skip the intermediate `Mat2` and multiply the two linear columns directly.

## Solution

This keeps the existing arithmetic order and affine-row assertion. Templates, generated code, tests, and benchmarks are updated.

Kept this to `Mat3A`: the same change made single-transform `Mat3` timing 4% slower.

## Benchmarks

On my Xeon E5-2696 v4:

| Measurement | Before | After | Change |
|---|---:|---:|---:|
| Single transform | 2.864 ns | 2.611 ns | -8.8% |
| 256-transform batch | 289.829 ns | 252.050 ns | -13.0% |
| Throughput | 883.28 M/s | 1015.67 M/s | +15.0% |
| Cycles/transform | 3.160 | 2.744 | -13.2% |
| Instructions/transform | 12.080 | 10.104 | -16.4% |
| Allocations | 0 | 0 | Unchanged |

Timing: Criterion on Windows. Counters: five alternating, core-pinned runs under WSL. The new loop performs one extra memory read per transform, but total bytes read and written stay unchanged.

## Testing and linting

- Six million comparisons passed: non-NaN outputs matched bit for bit; NaNs matched by classification.
- Regression tests cover signed zero, underflow, overflow, non-finite values, ignored translation, and the affine-row assertion.
- All nine feature configurations, release assertion tests, MSRV, and portable SIMD checks passed.
- Glam-specific Clippy, formatting, and matrix code-generation checks passed.
- Full workspace Clippy wasn't rerun due to the existing Windows CI-tool lint.

🤖 AI disclosure: I used Codex with GPT-6 Astra at xhigh reasoning effort to help identify the optimization and construct the benchmark. I have reviewed and understand the resulting implementation.

### DIFF
--- a/benches/mat3a.rs
+++ b/benches/mat3a.rs
@@ -3,8 +3,9 @@
 mod macros;
 mod support;
 
-use criterion::{criterion_group, criterion_main, Criterion};
-use glam::Mat3A;
+use criterion::{criterion_group, criterion_main, BenchmarkId, Criterion, Throughput};
+use glam::{Mat3A, Vec2};
+use std::hint::black_box;
 use std::ops::Mul;
 use support::*;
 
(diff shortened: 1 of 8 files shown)


bitshifter:
I think this optimisation is very layout dependent, rather than doing less work like the other changes I've merged, it's kinda doing the same work, but avoiding the Mat2 SSE2 implementation which is allowing the compiler to optimize the loop code better, in this case. It may not be uniformly true.

In any case, I am most likely going to switch Mat2 back to being entirely scalar  (see #830)  as the performance benefits of SIMD are inconsistent at best and SIMD Mat2 makes the type 16 byte aligned which means it doesn't pack nicely in types like Affine2. Since this optimisation is effectively just avoiding calling that SIMD code I think it will be a redundant change if Mat2 becomes entirely scalar, so closing this for now.

Issue #823, opened 2026-09-03
"Precise" lerp implementation is less precise for certain usecases


neocturne:
**Describe the bug**

#548 switched the lerp implemetation for `vec*` to a "precise" lerp. For many common usecases, this is actually *less* precise, as it produces rounding errors that are most noticable when the inputs are equal or close to each other.

**To Reproduce**

```rust
let a = Vec2::new(0.1, 0.1);
let result = a.lerp(a, 0.1);
println!("{a}");
println!("{result}");
```
prints:
```
[0.1, 0.1]
[0.099999994, 0.099999994]
```
So interpolating `a` with itself results in additional precision loss.

**Expected behavior**

Interpolating between two equal inputs should not result in a different output.

Note that the same issue also occurs if only one coordinate is equal, so in that case the lerp couldn't be skipped altogether.

**System information (please complete the following information):**

I noticed this with Bevy 0.19, which uses glam 0.32.1.

**Additional context**

I noticed this when using the interpolation feature of the Bevy physics engine avian. For this use case the new `vec*` behavior is undesirable: As my project uses nearest rather than linear texture filtering, the rounding errors can result in a entity that should be at rest to jitter between two pixel coordinates if its position is just at the boundary between two pixels.

Currently the behavior is also inconsistent between scalar floats and `vec*`/`quat`: `vec*` and `quat` have a `test_lerp_big_difference` test that checks the new behavior, while scalar floats do not have such a test and still use `a + (b - a)*s` for lerp.

I would argue that the old behavior is less surprising: `test_lerp_big_difference` is about interpolating between values that greatly differ in magnitude, but it is well-known that floating-point numbers can't deal with such values well, so a larger rounding error is to be expected in that case.

I think both approaches to lerp have merit, but I would prefer to have them as separate methods as you originally suggested in #278, as both have different rounding tradeoffs.

bitshifter:
Unfortunately there is no single lerp function to rule them all, both have different strengths and weaknesses.

One of the reasons I defaulted to the precise form is that it is the method that is specified by GLSL, WGSL and SPIR-V. HLSL and Metal allow both (up to the compiler implementation I guess). The issue you are running into is the lack of non-monotonic property of the precise form.

Therefore I don't plan to change the default but I will add a `lerp_monotonic` method that provides a monotonic implementation for use cases where that is required and try to document when you might want to choose monotonic.

[...]

PR #854, opened 2026-09-19
feat: add `lerp_monotonic` for float vector types

bitshifter:
# Objective

- Part of #823.

`lerp` keeps the form `self * (1 - s) + rhs * s` because that is the linear blend used by GLSL and WGSL. It is exact at `s == 1` even for inputs that differ greatly in magnitude, but it is not monotonic in `s` for nearly equal inputs and does not preserve equal inputs exactly, which causes jitter when interpolating between values that are at rest.

## Solution

- Add `lerp_monotonic` to `Vec2`/`Vec3`/`Vec3A`/`Vec4` and `DVec2`/`DVec3`/`DVec4`, using the monotonic form `self + (rhs - self) * s`.
- Use a fused multiply-add on SIMD back-ends where available: SSE2 via `m128_mul_add` (when the target has FMA) and NEON via `vfmaq_f32`.
- Leave `lerp` unchanged so existing callers keep the current rounding; #823 is addressed by opting into `lerp_monotonic`.
- Change `FloatExt::lerp` to the form used by the vector methods. This changes `remap`/`inverse_lerp` for infinite `t` from infinite to NaN.
- Document the form used by both methods and when to prefer each.

`lerp_monotonic` preserves equal inputs exactly and is monotonic in `s`. Because `rhs - self` is evaluated first, it is less accurate when the inputs differ greatly in magnitude and overflows for opposite-sign extremes.

Quaternions are unchanged. Normalisation already removes the equal-input jitter, and a monotone blend measured no benefit for `Quat`/`DQuat`.

## Tests

- `test_lerp_monotonic` covers the endpoints, exact preservation of equal inputs, monotonicity in `s`, and both tradeoffs above.
- Added `FloatExt::lerp` big-difference coverage and updated `test_remap`.


bitshifter:
I'm still bikeshedding what the best default option is for quat lerp and slerp fallback

